PR #10, opened 2022-07-08
Roam <> Matter

hunterclarke:


hunterclarke:
Review tasks:
- [ ] Migrate to new settings panel
- [ ] Return the config observer so it's properly removed
- [ ] Tear down the command palette command

panterarocks49 (approved):

PR #11, opened 2022-07-08
Dark theme toggle

8bitgentleman:
This extension adds a button to the topbar to allow toggling of custom CSS modes. It works by adding and removing the css class `.rm-dark-theme` from the document body.
Custom theme maintainers (or the adventurous) can add support for this in several ways but I recommend using `:root` variables.

```
:root{
	//REGULAR THEME COLOR VARIABLES
}
:root .rm-dark-theme {
	//DARK MODE COLOR THEME VARIABLES
}
```


8bitgentleman:
Thanks for the merge @panterarocks49 ! Looks like the example gif in my README isn't showing correctly in the Extensions page. Are they not supported
Screen Shot 2022-07-08 at 6 44 11 PM
?

panterarocks49:
I wasn't rendering html correctly, will be fixed in a few minutes

PR #12, opened 2022-07-08
html-table-import

8bitgentleman:
Right-Click menu plugin: Converts an HTML table copied to the clipboard into a Roam Table format. To use select the table text from a webpage ([this table used for testing](https://www.w3schools.com/html/html_tables.asp)) then in roam select the plugin from the right click block menu.


panterarocks49 (approved):
Cool!

PR #13, opened 2022-07-08
block-markdown-export

8bitgentleman:
# Block Markdown Export
Right-Click menu plugin: Adds an option to the right click menu to export a block and all its children as flat markdown. 

### Supports:

* Bold
* Italic
* Strikethrough
* Markdown tables
* Headings (H1, H2, H3)
* Block Refs (replaces the block ref with the actual referenced block contents)
* Blockquotes

### DIFF
--- a/extensions/8bitgentleman/block-markdown-export.json
+++ b/extensions/8bitgentleman/block-markdown-export.json
@@ -1,0 +1,10 @@
+{
+    "name": "Block Markdown Export",
+    "short_description": "Right-Click menu plugin: Exports a block and all its children as flat markdown",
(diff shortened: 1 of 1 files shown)


panterarocks49:
How is this different from Roam's export to flat markdown?

I get an error when trying to run this
image

Which points to here
image

8bitgentleman:
Ok wow I had no clue that if you zoomed into a block and exported to markdown it does exactly the same thing this plugin does (minus the export being 'flat' ie devoid of structure. Never noticed that, very cool! I'll close this plugin pull then

PR #21, opened 2022-07-15
Clear-sidebar button

8bitgentleman:
Adds a button to the top of the right sidebar to clear and close all blocks and pages in the sidebar. Not 100% sure about the styling but it looks find in vanilla roam.

### DIFF
--- a/extensions/8bitgentleman/clear-sidebar.json
+++ b/extensions/8bitgentleman/clear-sidebar.json
@@ -1,0 +1,11 @@
+{
+    "name": "Clear Right Sidebar",
+    "short_description": "Adds a button to the top of the right sidebar to clear and close all blocks and pages in the sidebar.",
+    "author": "Matt Vogel",
+    "tags": ["right sidebar", "button"],
+    "source_url": "https://github.com/8bitgentleman/roam-depot-clear-sidebar",
+    "source_repo": "https://github.com/8bitgentleman/roam-depot-clear-sidebar.git",
+    "source_commit": "4972e40f7085a297f67762c190dcf7f928f8a586",
+    "stripe_account": "acct_1LJFEYQmxalymEZL"
+
+}
--- a/extensions/8bitgentleman/clear-sidebar.json
+++ b/extensions/8bitgentleman/clear-sidebar.json
@@ -5,7 +5,7 @@
     "tags": ["right sidebar", "button"],
     "source_url": "https://github.com/8bitgentleman/roam-depot-clear-sidebar",
     "source_repo": "https://github.com/8bitgentleman/roam-depot-clear-sidebar.git",
-    "source_commit": "4972e40f7085a297f67762c190dcf7f928f8a586",
+    "source_commit": "0d50982067f182d087965482dfcb2dc24ac62682",
     "stripe_account": "acct_1LJFEYQmxalymEZL"
 
 }


baibhavbista:
Hey @8bitgentleman , some questions/feedback:

1. the button appears in the middle (slightly left offset) of my right sidebar, which seems to be different from your gif. Is that intentional?
image

2. You are handling "outline" and "block" types of windows, but we actually have 4 types (missing: "mentions", "graph"). Please handle all types
    - Relevant section of your code: https://github.com/8bitgentleman/roam-depot-clear-sidebar/commit/4972e40f7085a297f67762c190dcf7f928f8a586#diff-19a9a3efb94bf1b84229bda71ab0ef5badcdbf34ba3a0c9e4152a82c569a21e6R4-R18
    - Link to the types of windows in roamAlphaAPI docs: https://roamresearch.com/#/app/developer-documentation/page/YH7Res7Z2
    - Screenshot showcasing sample data for all types of sidebar windows    
      image

8bitgentleman:
Thanks for the review! Running into some trouble removing mentions from the sidebar and looking at the data types returned mine are actually different from yours. My mentions call return a 'page-uid' while yours return a 'mentions-uid' I checked for updates but I'm on the latest. What could be going on here?
image

baibhavbista:
Hey @8bitgentleman 
yeah it does seem like there's a bit of different data when we open mentions via different ways (via shift click on number of back refs vs via the roamAlphaAPI). Will look deeper into it later and see if one can be removed without breaking changes to extensions

In any case, cannot you do something like w['block-uid'] or w['page-uid'] or w['mentions-uid']. 
I think something like this should work for all cases (without the if statement)
```
window.roamAlphaAPI.ui.rightSidebar.removeWindow(
  {"window": 
      {"type": w['type'], 
        "block-uid": w['block-uid'] || w['page-uid'] || w['mentions-uid']}
  }
)
```

baibhavbista (commented):
https://github.com/Roam-Research/roam-depot/pull/21#issuecomment-1190463856

8bitgentleman:
@baibhavbista  Thanks for the review, that's much more straightforward. Are you seeing any irregularities around removing mentions via the alphaAPI? This is what I'm seeing
Kapture 2022-07-20 at 12 49 15

HD vid 👇
https://user-images.githubusercontent.com/4028391/180040480-223685fb-63af-4153-ac6f-51fadc0b31b9.mp4

baibhavbista:
Hey @8bitgentleman , yes I was able to reproduce the bug
It was an issue on our part
Will release a fix soon. Can you update roam and check in a couple of hours?

8bitgentleman:
@baibhavbista I tested on that deployment. Mentions remove as expected but now I'm running into the same issue with the `graph` type `Promise {<fulfilled>:null}` when trying to remove via API

baibhavbista:
Hey @8bitgentleman thanks for the check, I think I've fixed it now
Can you check in the following deploymenbt? http://relemma-cjo9ipxg8.roamresearch.com/

8bitgentleman:
@baibhavbista  that works as expected! Just cleaned up the code per your suggestion. 
Also I've chosen to not remove sidebar items that are pinned. That seems like expected behavior

baibhavbista:
Merged the PR with the above fix, will merge this extension after sometime

PR #25, opened 2022-07-18
Show favicon

paulovieira:


panterarocks49:
Can you remove the styles in `onunload`? 
One of the observers is also not being properly removed, https://github.com/paulovieira/roam-show-favicon/blob/c4875ed5204fad289066783f91d255f2b6ee1131/extension.js#L45 overwrites the first observer that was set and then only one of them is removed

Users should be able to uninstall the extension and it removes the functionality without reloading the page

Thanks!

paulovieira:
> Can you remove the styles in `onunload`? 

Deleted the `extension.css` file (it was empty, this extension doesn't need any css styles).

>One of the observers is also not being properly removed, https://github.com/paulovieira/roam-show-favicon/blob/c4875ed5204fad289066783f91d255f2b6ee1131/extension.js#L45 overwrites the first observer that was set and then only one of them is removed

Good catch! The 2 observer instances are now stored in `internals.observers` (in the `startObserver` function, which is called in `onload`). They are then stopped in `stopObserver`, which is called in `onunload`.

> Users should be able to uninstall the extension and it removes the functionality without reloading the page

This is now implemented in `removeFavicon`, which is called in `stopObserver`.

Thanks for the feedback.

PR #26, opened 2022-07-19
ptn-roam-depot

scottrblock:
my repo (https://github.com/phonetonote/ptn-roam-depot) has the extension built, but can also be rebuilt with the `build.sh` script.

the code is mostly the existing phone-to-roam-client (https://github.com/phonetonote/phone-to-roam-client/), but I removed old dependencies, and made an onboarding flow for new and existing users.

### DIFF
--- a/extensions/phonetonote/phonetonote.json
+++ b/extensions/phonetonote/phonetonote.json
@@ -5,6 +5,6 @@
   "tags": ["phonetonote", "ptn"],
   "source_url": "https://github.com/phonetonote/ptn-roam-depot",
   "source_repo": "https://github.com/phonetonote/ptn-roam-depot.git",
-  "source_commit": "76b9f47ef05434dfec82dbbb35bd31b0baea78f7",
+  "source_commit": "19d2af2bc976d0fe292bebf53f5aaf69a4a45728",
   "stripe_account": "acct_1LKwanQVF0QOKIg5"
 }
--- a/extensions/phonetonote/phonetonote.json
+++ b/extensions/phonetonote/phonetonote.json
@@ -5,6 +5,6 @@
   "tags": ["phonetonote", "ptn"],
   "source_url": "https://github.com/phonetonote/ptn-roam-depot",
   "source_repo": "https://github.com/phonetonote/ptn-roam-depot.git",
-  "source_commit": "19d2af2bc976d0fe292bebf53f5aaf69a4a45728",
+  "source_commit": "39ee1ccd1e60f71fb51221f2e1c965cf68c04173",
   "stripe_account": "acct_1LKwanQVF0QOKIg5"
 }
--- a/extensions/phonetonote/phonetonote.json
+++ b/extensions/phonetonote/phonetonote.json
@@ -5,6 +5,6 @@
   "tags": ["phonetonote", "ptn"],
   "source_url": "https://github.com/phonetonote/ptn-roam-depot",
   "source_repo": "https://github.com/phonetonote/ptn-roam-depot.git",
-  "source_commit": "39ee1ccd1e60f71fb51221f2e1c965cf68c04173",
+  "source_commit": "64d85e9877a26db73ed7298640f960e979574f04",
   "stripe_account": "acct_1LKwanQVF0QOKIg5"
 }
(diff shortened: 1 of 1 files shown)


panterarocks49:
I didn't get time to fully review today but did take a glance. Could you update the readme with a better description? It will be displayed inside of the marketplace

It's not clear from the description or the readme that I would be able to text notes (or telegram, facebook etc) to my Roam graph with this extension

also don't need to have "ptn-roam-depot" in the readme since the full name will be displayed above inside of Roam

scottrblock:
thanks @panterarocks49, I just updated the README. let me know if i need to update anything on this PR

panterarocks49:
Could you update from roam-client to roamjs-components? roam-client is marked as depreciated

panterarocks49:
The first load of phonetonote results in this

image

I reloaded and it fixed the error

panterarocks49:
Summary of review this far
Blockers
1. upgrading from roam-client to roamjs-components, there looks like a number of security vulnerabilities coming from packages required in roam-client and it is marked as depreciated
2. initial sign in results in "ptn is loading" have to reload error

Nice to have fixes but not blocking
1. updating my settings from the config panel isn't changing the settings for me, I have to update the input and reload for the changes to apply
2. multiple word hashtags don't work, ei, "Quick Capture" results in "#Quick Capture" instead of "#[[Quick Capture]]"

it looks like indexpages is not used anymore, is that correct?

Is bugsnag necessary? 

Is it possible to set it up with multiple graphs or is it tied to only one?

I think before we merge this I'm going to add a key to the json file that this extension sends user data (your email) to a third party service

panterarocks49:
also, would be nice to remove the code that isn't being used anymore

scottrblock:
Thanks @panterarocks49, I just pushed some commits to fix most of this:

> upgrading from roam-client to roamjs-components,
done

> initial sign in results in "ptn is loading" have to reload error
fixed and tested e2e a bunch, it should be good. let me know if you can reproduce any issues. I can delete your ptnkey on our backend for a more thorough test, just DM me your current key.

> updating my settings from the config panel isn't changing the settings for me
fixed

> multiple word hashtags don't work, ei, "Quick Capture" results in "#Quick Capture" instead of "#[[Quick Capture]]"
created an issue for this, will get to soon but doesn't seem blocking - https://github.com/phonetonote/ptn-roam-depot/issues/1. 

> it looks like indexpages is not used anymore, is that correct?
this should all be removed, where are you seeing it?

> Is bugsnag necessary?
removed it

in general let me know if you see any extra code still being included, i tried to slim things down. appreciate the thorough review!

panterarocks49:
> this should all be removed, where are you seeing it?

Sorry, somehow I think I got messed up and was looking at a past version when I wrote that

panterarocks49:
I am still getting the error on first sign in (I tried with a second Roam account I have)
also noticed these errors in the console
image

panterarocks49:
video of the bug happening

https://www.loom.com/share/59d5722d09054deea2e00214a1b379ea

scottrblock:
thanks @panterarocks49, I just pushed some updates to my backend to fix this. You can retest with a new account, or do something like this in `ptn-roam-depot.tsx`:

image

tldr is it was a race condition, but i was able to completely remove the root cause, so should be good to go.

scottrblock:
hey @panterarocks49 i'm finishing up the very last edge case i can think of, where if someone is already logged into my ptn dashboard, but they somehow end up with a new ptn account from roam depot (it would likely be a second account scenario, but it is possible for me to not be able to reconcile a new user coming in from roam depot. It's also what you're hitting when testing with multiple accounts. And possibly other people have other accounts). Hoping to have this done some time this afternoon, it's a bit tricky to track down/test, but I'm making progress.

scottrblock:
hey @panterarocks49 everything is ready for launch now. the link to the dashboard correctly handles new users, existing users, and rare edge cases of the user already being logged into phonetonote under a different account that I can't reconcile. I spent the afternoon trying to break different edge cases and can confidently say it's well tested and ready. feel free to give it another spin and merge!

panterarocks49:
Thank you for all of your work on this! 
The UI is fixed for me now, but when I clicked it, it took me to phonetonote and displayed "something went wrong", refreshing fixed the error. This is at least a bit more obvious that I need to refresh to fix it though, and I'd guess it's probably just a race condition on your backend so maybe people aren't even going to experience that if they take a little bit longer to hit the button.
Merging it now

PR #29, opened 2022-07-19
Added Roam CSS System III

rcvd:


panterarocks49:
Hey we plan to add a different option for themes later that will have a separate review process and user interface, I'll accept this PR since we don't have that yet and because it has it's own configuration, but keep in mind you may need to migrate later.
I will need you to change your code though and include the css in the repository, we don't allow dynamically loading css because of attacks like these https://css-tricks.com/css-keylogger/

rcvd:
I understand. My CSS System currently uses more than 120 CSS files for the six themes, modes, and modules. If I put them into the plugin repository, how can I then selectively load them from the plugin? I can't combine it all into a single extension.css.

Regarding a potential migration: I would like to invest more time into the plugin to make various additional things customizable (Heading Size, Font Families, Accent Colors) via UI - my CSS System already exposes more than 1,700 easily editable variables for all of this. Adding an option in the plugin is just a field setting a CSS property, so this could quickly grow. But I don't want to spend all this effort if it is already clear that I must migrate and (most possibly) lose this functionality. What are your thoughts about this?

panterarocks49:
Does that mean that loading a theme could potentially make 120 network requests to fetch all of the required css?

One approach to putting multiple themes into a single file is nesting all the css inside of a top class, and then adding that class to the root when the theme is turned on. You'd also need to compile all of the css into a single file with something like less or sass.

```
.rcvd-bear-theme {
   .bear-theme-class {
   }
   ... more bear theme variables or classes
}
```
and then toggle the tag on the body
```
document.body.classList.toggle("rcvd-bear-theme")
```

Another approach would be to use a style loader with webpack, https://webpack.js.org/loaders/style-loader/. I am not as familiar with this approach but I think it inlines the css into your js code

On the migration in the future, we would still require that your theme is in a single css file. I think we would also allow you to have a js file to add a customization panel if you want.
The main difference I was planning with themes is that it only makes sense to have one theme applied at a time, so Roam would provide the UI for switching themes, which would mean each of these themes would have to be submitted individually.

rcvd:
Thanks for your suggestions.

No, a single theme makes only a few calls. But six themes are included with three modes each (light, dark, auto (follows the current OS setting)). 

I have my system.css that provides variables for all of Roam's relevant classes and attributes (more than 1,800 attributes for more than 400 objects at the moment) and will always be loaded. 

Example for the Danger-Button:
```
.bp3-button.bp3-intent-danger {
    background-color: 	        var(--bc-button__danger);
    border-top: 		var(--bt-button__danger);
    border-right: 		var(--br-button__danger);
    border-bottom: 		var(--bb-button__danger);
    border-left: 		var(--bl-button__danger);
    border-radius: 		var(--bd-button__danger);
    box-shadow: 		var(--sd-button__danger);
    --webkit-box-shadow: 	var(--sd-button__danger);
    color: 			var(--co-button__danger);
}
```

Each theme contains the variable settings for all the attributes, and if a user wants, he can change all of the variables afterward. For example, if a user wants a violet danger button, all he has to do is add a CSS block:

```
:root {
    --bc-button__danger: violet;
}
```

**Customizations of my CSS System are widely used in the roam community but are often still found to be complicated even with my documentation (https://bit.ly/roam-css-sys). These customizations are the main reason for me making a plugin out of it.** 

Additionally, most themes load a custom color scheme, new fonts, and layout fixes (bullet alignment, for example). If you count everything together, you'll get these almost 120 files. Since the introduction of http2, combining CSS files is not recommended anymore because parallel transfers can accelerate the load times. But a single theme won't load more than seven files.

If you are curious, you can see all the relevant files in my repository (https://github.com/rcvd/roam-css-system). I showed this system to Conor last year. 

I've been thinking a lot about the design of my CSS system while developing it over the last two years to keep it maintainable, customizable and powerful. It took me a lot of my spare time (but made a lot of fun). I was hoping to easily make a plugin out of it and use the UI for configuration. But it looks like it's getting much more complicated than I thought.

rcvd:
@panterarocks49: I'm still trying to get my whole CSS System into the Roam Depot Plugin. Because we are not allowed to load themes remotely, we also can't load fonts remotely. I, therefore, also have to embed the additional fonts into the extension.js, which can be done base64 encoded. BUT the extension.js will get rather large, I expect around 10 MB, with 99% being encoded fonts and minimized CSS (I'll try to put all that stuff below the real code so you can check it more easily). Are you okay with this?

8bitgentleman:
@rcvd you may be interested in my [Dark Theme Toggle](https://github.com/8bitgentleman/roam-depo-dark-toggle) extension which was recently merged to Roam Depot. I've actually used that plugin with some of your themes modified to support. Perhaps you could include support for that extension in your themes in the future so they could work symbiotically?

panterarocks49:
@rcvd I am not aware of a security risk from loading fonts dynamically, but it would be nice if the themes worked offline, so I think putting the font into the extension.js would be the best approach (we cache extension.js and extension.css)
I do need to be able to review the css code, so submitting only an extension.js with minified css in it isn't ideal, preferably you would have a build process that minifies the css and inserts it into the js file (or extension.css)

rcvd:
> preferably you would have a build process that minifies the css and inserts it into the js file (or extension.css)

@panterarocks49: This is where I'm now:
- I provide all the un-minified CSS files for my CSS System (giving Roams class attributes variables), all of my themes (setting these variables and fixing small layout glitches), and the modules (giving addition functionality like centering the search dialog) within the repository
- I also provide a small javascript file for the extension **without** the embedded CSS data, that is meant to load the themes and modules
- I have a build.sh that first minifies the CSS files and after that appends these to a new function to the javascript file that then will compile to extension.js

Could you please confirm that I'm on the right track now?

panterarocks49:
@rcvd yes, that sounds good to me.

panterarocks49:
I am closing this PR for now because it is stale, please reopen if / when you decide to make changes

PR #33, opened 2022-07-20
Jay Bienvenu's "Snapshot" command

jbnv:


panterarocks49 (changes requested):

panterarocks49:
hmm, our json parser is still failing, could you try removing the trailing comma?

baibhavbista:
Hey @jbnv , 
From reading the README, I did not understand what the extension was meant to do. Could you add some context and maybe use cases?
Based upon my experiments with the extension, I think the main goal of the extension seems to be to export the current state of the opened roam page, so that it can be printed out (as pdf or on paper). Is that correct? What do you use the extension for?

Some other remarks:
1. When running the snapshot command from the command palette, it gives me a html file with no margin on the left (to the level that the top most level of blocks' bullets are cut in half). Can you look at that?
Here's a screenshot: 
image

2. A note regarding the code: I think you should be able to keep only https://github.com/jbnv/roam-snapshot/blob/6a5f7131ac44f7dcacb5d2068f3844a3fd534562/extension.js#L138-L147 in the onload function and move all of the other functions out. 

3. In the onunload function, you should handle any cleanup. In this case, it makes sense to remove the command that was added to the commandPallete. Here's the documentation for roamAlphaAPI.ui.commandPalette.removeCommand: https://roamresearch.com/#/app/developer-documentation/page/orAx-_9t5

panterarocks49:
@jbnv ping on the above comments

panterarocks49:
Closing this since there is no activity, please reopen when you address the comments above

PR #35, opened 2022-07-21
hide-left-sidebar-buttons

8bitgentleman:
Plugin to hide/show the individual buttons in the left sidebar based on a request in the slack channel. Since new plugins have been adding buttons to the sidebar (and more are likely to come) I decided to go the JS route instead of just using CSS `nth-child()`

  </img>

### DIFF
--- a/extensions/8bitgentleman/hide-left-sidebar-buttons.json
+++ b/extensions/8bitgentleman/hide-left-sidebar-buttons.json
@@ -1,0 +1,10 @@
+{
+    "name": "Hide/Show Left Sidebar Buttons",
+    "short_description": "Roam Research plugin to hide/show the individual buttons in the left sidebar",
+    "author": "Matt Vogel",
+    "tags": ["left sidebar"],
+    "source_url": "https://github.com/8bitgentleman/hide-left-sidebar-buttons",
+    "source_repo": "https://github.com/8bitgentleman/hide-left-sidebar-buttons.git",
+    "source_commit": "ad00be5f53a9a71e1b4895f54a71aaa21154bcce",
+    "stripe_account": "acct_1LJFEYQmxalymEZL"
+  }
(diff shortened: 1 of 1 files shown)


panterarocks49:
Hey something is wrong with the link above, can you double check the github repo / commit?

8bitgentleman:
Whoops, sorry about that @panterarocks49 . Should be fixed now

baibhavbista:
Hey @8bitgentleman, I'm curious, you mention "Since new plugins have been adding buttons to the sidebar" in the comment above. Currently this is only for the official left sidebar buttons, right? How will this deal with those added by plugins/extensions?

8bitgentleman:
@baibhavbista Correct this is only for the official left sidebar buttons right now. I figured if another plugin adds a button to the left sidebar then a user would just remove that plugin to remove the associated left sidebar button. 

(only the Random Button, and Bitcoin Price Tracker are in Roam Depot now but there are other community plugins with left sidebar buttons that have yet to be ported to Roam Depot)

baibhavbista:
Hey @8bitgentleman, yeah that makes sense

I only have some small nitpicks:
1. if you do it the way shown in the right, if there are other extensions styling those elements, they will persist across hide and show cycle
image

2. I think you can keep the console logs for load/unload, but can you remove the ones that print all the classes?
image

8bitgentleman:
@baibhavbista Makes sense, I have updated with those cleanups

baibhavbista:
@8bitgentleman thanks for the updates

Will merge this extension now

baibhavbista:
Hey @8bitgentleman 
Sorry I had to revert the PR with your extension

It turns out that for the ids in settings in the panel, they must be non-empty strings and can't contain ".", "#", "$", "[", or "]"
(required by the backend where we're storing the extension-level settings)

More details are in the revert PR: https://github.com/Roam-Research/roam-depot/pull/46

Can you please fix that issue and make a new PR?

PR #46, opened 2022-07-26
Revert "hide-left-sidebar-buttons"

baibhavbista:
Reverts Roam-Research/roam-depot#35

Issue was that in the settings panel, the id was something like ".bp3-icon-calendar". 
However, they must be non-empty strings and can't contain ".", "#", "$", "[", or "]"

Added a note for the same in https://roamresearch.com/#/app/developer-documentation/page/y31lhjIqU
image


baibhavbista:
@panterarocks49 potential issue when dealing with settings panel
I think it makes sense to have a check for this in the code for settings panel (since unless we manually check for it every time, we would not know issue until extension is in production)

PR #47, opened 2022-07-26
Left Sidebar Buttons: fix illegal char in settings id

8bitgentleman:
as detailed in #46 setting ids be non-empty strings and can't contain ".", "#", "$", "[", or "]"


baibhavbista:
looks good. Matt has changed the ids to not have the "."s that were causing the issues

PR #53, opened 2022-07-29
Pull request for Universal Quick Capture extension

mlava:


tonsky:
Hi! Thanks for submitting your extensions to Roam Depot.

It would be much easier for review and we can get some extensions in faster if you can create a separate PR, one per extension.

So far I have some notes on the first one (mlava/roam-json-webhooks):

- JSON file name should match github repo name exactly (`extensions/mlava/json-webhooks.json` → `extensions/mlava/roam-json-webhooks.json`)
- You seem to be depend on jQuery (`$.ajax`), which is not bundled with Roam. Change to `fetch`?
- If Webhook Address or Delimiter are not specified, extension fails silently. Maybe show a message?
- README doesn’t make it clear what data exactly the extension sends to the webhook. Maybe start with an example of Roam page, where in the page the extension could be invoked and what resulting JSON will be sent?
- Concatenating strings to build a query is a bad practice that might lead to security exploits. Consider using query paramenters instead:

```
const startBlock = await window.roamAlphaAPI.ui.getFocusedBlock()?.["block-uid"];
var apiCall = `[:find ?ancestor (pull ?block [*])
                :in $ ?b
                :where [?ancestor :block/uid ?b]
                       [?ancestor :block/string]
                       [?block :block/parents ?ancestor]]`;
var childBlocks = await window.roamAlphaAPI.q(apiCall, startBlock);
```

Notice `:in` block in query that binds `?b` and extra parameter passed to `.q` with block id

tonsky:
mlava/roam-open-close-children:

- We have a cmd+up/cmd+down keyboard shortcuts to do just that. Why is extension needed?
- JSON file name should match github repo name exactly (extensions/mlava/json-webhooks.json → extensions/mlava/roam-json-webhooks.json)
- Same note on using query parameters
- You seem to be over-querying (querying too much and then not using it). Consider simpler query:

```
let q = `[:find ?string .
          :in $ ?b
          :where [?page :block/uid ?b]
                 [?page :block/string ?string]`;
```

tonsky:
mlava/roam-random-quote:

- Same note on JSON file name (random-quote.json → roam-random-quote.json)
- Same note on jQuery usage (replace with `fetch`)
- This seems unnecessary?
```
var jsonQuotes = JSON.stringify(response);
var quote = JSON.parse(jsonQuotes);
```
- Should you unregister smartblocks command in unload?

tonsky:
mlava/roam-send-readwise:

- Same note on JSON file name (send-readwise.json → roam-send-readwise.json)
- Same note on jQuery usage (replace with fetch)
- Same note with over-querying and string concatenation (replace with query parameters, only query what you use)

tonsky:
mlava/roam-universal-quick-capture:

- Name is misleading, rename to e.g. import-todoist-quick-capture
- Same note on JSON file name (must match repo name)
- Same note with handling missing/not configured yet settings (show a message)
- Same note on jQuery usage (replace with fetch)
- Same note with over-querying and string concatenation (replace with query parameters, only query what you use)

tonsky:
mlava/roam-weathercards:

- dayjs is not defined
- errors are not handled, including missing info in settings
- Temperatures look a bit too high

Screenshot 2022-08-04 at 20 02 11

Otherwise looks great!

tonsky:
mlava/roam-word-count:

- Looks fine :) Unfortunately, I can’t merge it separately unless you create a separate PR for it

mlava:
@tonsky 
Thank you so much for reviewing each of these extensions and the detailed feedback. I will start work on improving the code/submissions and resubmit as separate PRs. Cheers

mlava:
> Hi! Thanks for submitting your extensions to Roam Depot.
> 
> It would be much easier for review and we can get some extensions in faster if you can create a separate PR, one per extension.

Thanks @tonsky.
I am trying to create a separate PR for each extension, but am having trouble figuring out how to do it. I have forked the roam-depot to my github. I have committed updates to the word-count.json file to a new branch and then tried to submit a PR.
The PR still lists all the other commits I've made to other extensions in my fork.
Can you advise how to create a proper separate PR within github? Sorry for noob issue.

mlava:
> mlava/roam-open-close-children:
> 
> * We have a cmd+up/cmd+down keyboard shortcuts to do just that. Why is extension needed?

@tonsky 
I had forgotten. Won't resubmit.
Cheers!

tonsky:
Ha, interesting! I’ve never thought about that, since it’s a single repo and you probably want just one fork. What if you create a branch per PR? I think when you make a PR it lets you select source and destination branches

mlava:
> mlava/roam-weathercards:
> 
> * Same note on JSON file name (must match repo name)
> * dayjs is not defined
> * errors are not handled, including missing info in settings
> * Temperatures look a bit too high
> 
> Screenshot 2022-08-04 at 20 02 11
> 

Weird. I get this for Berlin, which looks correct:

image

On dayjs - it is available within Roam Research isn't it? I didn't realise I needed to define it?

mlava:
> Ha, interesting! I’ve never thought about that, since it’s a single repo and you probably want just one fork. What if you create a branch per PR? I think when you make a PR it lets you select source and destination branches

Here's a loom: https://www.loom.com/share/9a5f41c623cf44ff9bd40c33b7dcc44f

I committed the updates to word-count.json to a new branch and then tried to create a PR to roam-depot main.

You can see it still lists all of my other non-merged commits. Any suggestions?

tonsky:
Yeah git might be tricky, and github adds a layer of compexity on top of that. But don’t worry, we can make it work. First we need to get your repo into the shape that will allow you to make separate PRs. This is (roughly) a state of your repo right now:

Frame 1

And this is where we want it to be:

Frame 2

Basically we want each JSON to be in a separate branch that is branched off Roam-Research:main, with no other commits in that branch.

I also see that you have a few temporary commits in between where you probably was trying things out, which makes history a little harder to work with. I don’t think we necessary need to preserve that.

So this is what I propose:

```
git branch word-count-2 2caf9aba1489fb354d3ec17f9ba8d06df1d78893
git checkout word-count-2
git checkout main -- extensions/mlava/word-count.json
git add extensions/mlava/word-count.json
git commit -m "Add mlava/word-count.json"
git push
```

Here 2caf9aba1489fb354d3ec17f9ba8d06df1d78893 is the top commit of upstream `Roam-Research:main`. This will essentially create a situation like this:

Frame 3

From there, you should be able to create PR from `word-count-2` and it should contain just one commit with just one file `extensions/mlava/word-count.json` in its latest state.

If you then want to update any of the extensions, go to the corresponding branch (not main!) and make another commit to that file there. If PR was not yet accepted, it’ll automatically update. If it was accepted, you’ll have to create a new one.

Repeat for each extension and you should be good to go! Sorry it sounds complicated but I don’t think we can make it easier with git/github and the history you already have in your repo.

mlava:
@tonsky 
re https://github.com/Roam-Research/roam-depot/pull/53#issuecomment-1205996664

I thought dayjs was available within RR and didn't need to be defined? Is that incorrect?

tonsky:
> I thought dayjs was available within RR and didn't need to be defined? Is that incorrect?

Dayjs is not provided anywhere in your code and it’s not bundled with Roam itself

PR #54, opened 2022-07-29
Update phonetonote to v0.0.2

scottrblock:
There were previously issues where the extension was sometimes sending two requests to create a new account or for a sign in token instead of one.

The latest commit attempts to fix that by generally cleaning up the hooks and better managing settings. 

@panterarocks49 if there's any way you could test this like you did the first version, that would be great, as I can't reproduce any of this locally.


panterarocks49:
Issue is fixed for me

PR #55, opened 2022-07-30
FIX Left Sidebar Button CSS Selectors

8bitgentleman:
Y'all were tricky and changed the CSS & HTML structure on me. With Roam's new CSS before this fix the entire left sidebar would disappear. This should fix it


baibhavbista:
Oh sorry @8bitgentleman 
Will merge this PR right away

PR #56, opened 2022-07-30
Add Anki/Roam sync extension.

chronologos:


panterarocks49 (changes requested):
Thanks for submitting to Roam Depot!

There are a couple of small things you need to fix
 - Your extension needs a `build.sh` since it is compiled, and don't commit the extension.js to the repository, we will compile it on our end.
 
 - https://github.com/chronologos/Fabricius/blob/f3c5a48d810344042b412da368af59e3c4e46842/src/main.ts#L150
all setup and teardown should be done inside `onload` and `onunload`, you need to remove these lines. This is done so users can enable and disable extensions without reloading the page.
 - I was not able to get it to sync to anki, this might be just me not knowing how to use anki though. I copy pasted the two example you have, then clicked sync. When I go to anki "browse" I cannot find them though. This is what prints out, any ideas what could be going wrong?
image


Some other (non blocking) things it would be nice to have
 - Would be nice to link to anki connect plugin for anki so I know to install it
 - Would be nice to have a config page to change variables, [example](https://github.com/panterarocks49/settings-panel-example/blob/main/extension.js)
 - https://github.com/chronologos/Fabricius/blob/f3c5a48d810344042b412da368af59e3c4e46842/src/roam.ts#L135
:block/title isn't a valid attribute, either should be block/string for blocks or node/title for pages, although it doesn't look like you are using this anywhere critical

chronologos:
Thanks for the detailed feedback. I believe I have addressed all of them. For the cards not appearing in Anki, I typically search for `added:1` in the Browse window after running a sync. Did that work for you? Could you try with text in Roam that looks like https://github.com/chronologos/Fabricius#recommended-use ?

panterarocks49:
Something is wrong with the imports as well I think, but haven't used esbuild myself so not exactly sure how to fix it.
React, react-dom, and blueprint should not be included as dependencies, they should be reusing Roam's which are exported on the window object. In webpack you would do it like this https://github.com/panterarocks49/autotag/blob/main/webpack.config.js#L2. Could you fix this so you reuse Roam's blueprint and react?

Also, I think you have the wrong commit attached to this PR, it's not the latest on master

panterarocks49:
Thanks for fixing the other stuff, but I'm still not able to see anything in anki

Here's a short video of me trying to sync https://www.loom.com/share/291859e5b6f74b3cbfaabd89092de2a4

chronologos:
Thanks; I had a wrong comment in the code so I improved the logging messages. Basically AnkiConnect wasn't able to add the notes (and this could happen for multiple reasons). From the video I couldn't tell, but it's likely the deck name or model name doesn't exist (can be changed in settings).  I made some fixes to the settings parsing to fix default values too.

I was able to re-use Roam's React and ReactDOM via webpack as you suggested. Couldn't get blueprintjs to work though; I keep getting an `Uncaught (in promise) TypeError: Cannot read property 'Intent' of undefined` error at runtime. For now, I've included it as a dependency via webpack. Would you have any pointers for fixing this?

panterarocks49:
For blueprint, take a look at this https://github.com/dvargas92495/roamjs-scripts/blob/main/src/common/getBaseConfig.ts#L57

panterarocks49:
Ok now I am able to get it to sync, I guess I totally missed these instructions the first time
image

But I am also getting errors, one alert says failed to sync, the other says synced successfully
image

chronologos:
thanks once again @panterarocks49 ; I tried debugging this more but to no avail. As far as I can tell, the AnkiConnect plugin does also tend to flake out sometimes and drop the http connection. This typically happens when Anki is inactive. Fortunately, the solution is usually just to retry the sync. I have included more debug logs in case we want to try to track this down further.

panterarocks49:
Ok, I'm going to go ahead and release this

panterarocks49 (approved):

PR #57, opened 2022-07-31
Add roamist

kdnk:


panterarocks49 (changes requested):
Thanks for submitting to Roam Depot!

I get an error when I try to run it

image

I think this is because the roamist page doesn't exist, running the extension a second time fixes it, could you please fix this?

https://github.com/kdnk/roamist/blob/bbfc716b2d9783bf206f07d66a99c6eba51e2b30/src/js/init.tsx#L92
if you are fully using our config panel now can you remove this

You don't need to change this if you don't want, but you can use our function now https://github.com/kdnk/roamist/blob/bbfc716b2d9783bf206f07d66a99c6eba51e2b30/src/js/utils/convert-date-to-roam.ts#L2
see https://roamresearch.com/#/app/developer-documentation/page/G9TZzJ0oO 

Could you please clarify the docs on what to do when installing from roam depot? the readme includes a bunch of stuff about how to install via roam/js

kdnk:
@panterarocks49 

> I think this is because the roamist page doesn't exist, running the extension a second time fixes it, could you please fix this?
> 
> [kdnk/roamist@bbfc716/src/js/init.tsx#L92](https://github.com/kdnk/roamist/blob/bbfc716b2d9783bf206f07d66a99c6eba51e2b30/src/js/init.tsx#L92)
> if you are fully using our config panel now can you remove this


Thanks! That's right. In the current implementation, I guess I still need to have a specific setting page to generate some blocks for SmartBlocks. I couldn't find any ways to do that only with roam-depot panel. So I just fix this issue and keep using `createConfigObserver`.  I'll find a way to do configuration only with roam-depot panel.


> You don't need to change this if you don't want, but you can use our function now [kdnk/roamist@bbfc716/src/js/utils/convert-date-to-roam.ts#L2](https://github.com/kdnk/roamist/blob/bbfc716b2d9783bf206f07d66a99c6eba51e2b30/src/js/utils/convert-date-to-roam.ts#L2)
> see [roamresearch.com/#/app/developer-documentation/page/G9TZzJ0oO](https://roamresearch.com/#/app/developer-documentation/page/G9TZzJ0oO)

I didn't know that. I've changed my code using `dateToPageTitle `👍

panterarocks49 (approved):
Thanks for fixing those things, looks good now

PR #58, opened 2022-08-02
Add Readwise Daily Quote

TedisAgolli:
Show a quote from your Readwise account in your Daily Note page
image

### DIFF
--- a/extensions/readwiseDailyQuote/readwiseDailyQuote.json
+++ b/extensions/readwiseDailyQuote/readwiseDailyQuote.json
@@ -1,0 +1,12 @@
+{
+    "name": "Readwise Daily Quote",
+    "short_description": "Show a quote from your Readwise account in your Daily Note page",
+    "author": "Tedis Agolli",
+    "tags": [
+        "readwise",
+        "quote"
+    ],
+    "source_url": "https://github.com/TedisAgolli/roam-readwise-daily-quote",
+    "source_repo": "https://github.com/TedisAgolli/roam-readwise-daily-quote.git",
+    "source_commit": "fec4f58b74c0f6ccc086cb01cf4fb773160837ab"
+}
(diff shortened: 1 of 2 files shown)


TedisAgolli:
`      uid: window.roamAlphaAPI.util.generateUID().slice(0, 3) + "-" + "quote",
`
I'm not sure if messing with UIDs like this could cause any problems. I want to know if this page already contains a quote. I could insert that data in the settings but then I'd have to worry about cleaning it up so I don't keep adding entries for each day. Not that complicated, but I think more error-prone than what I'm doing now.

panterarocks49 (changes requested):
Hey thanks for submitting to Roam Depot!

Main thing you need to change is to make the folder and file name match the githubrepo, `TedisAgolli/roam-readwise-daily-quote.json`

Also, if I leave this running for multiple days, will it add a new note on the next day? or do I have to refresh Roam? I can't see anything in the code that would indicate that it will add a new one when the next day is reached

> uid: window.roamAlphaAPI.util.generateUID().slice(0, 3) + "-" + "quote",
[code](https://github.com/TedisAgolli/roam-readwise-daily-quote/blob/master/extension.js#L136)
I'm not sure if messing with UIDs like this could cause any problems. I want to know if this page already contains a quote. I could insert that data in the settings but then I'd have to worry about cleaning it up so I don't keep adding entries for each day. Not that complicated, but I think more error-prone than what I'm doing now.

It shouldn't cause any issues in vanilla Roam, but other extensions might bug out with that, but I'm not sure. Probably fine to leave as is.

TedisAgolli:
I thought it would be enough for it to be done onLoad, assuming people would reload Roam at least every day. It makes sense that you might just leave Roam open for days. I'll need to set up a timer then. I'll do that and update the folder/file name. 

Thanks for the feedback :)

panterarocks49:
The timeout should be cleared `onunload` as per our guidelines for extensions. If a user uninstalls an extension it should stop doing everything it was doing. Also if an update happens `onunload` is called for the previous version before `onload` is called for the new version, so clearing the interval will stop 2 versions from running at the same time.

panterarocks49 (approved):
Thanks for fixing!

PR #60, opened 2022-08-03
initial commit shodty breadcrumbs

shodty:
first attempt at this, so hope this code is up to snuff! please advise if not.


shodty:
Realizing I developed this with Roam JS and not using the Developer Extensions inside Roam, so it isn't properly working yet as a Roam Depot extension. Working to fix now, so closing this until it works properly.

PR #61, opened 2022-08-03
Roam breadcrumbs initial commit

shodty:
Fixed this to work as an extension. seems to be working as intended in developer mode, so I hope that means it's good to go!


panterarocks49 (changes requested):
I don't seem to have the same icon you do in your example picture, are you relying on an external theme's css for it?

also I get some not so nice overflow when I open the sidebar, are you able to put the element inline so that it adjusts with flexbox?

image

panterarocks49:
@shodty ping on the above questions

panterarocks49:
Closing this since I didn't get a response, please reopen if you still want to submit this

PR #63, opened 2022-08-03
Smartblocks: Version 2022-08-03-21-21

dvargas92495:


dvargas92495:
@panterarocks49 @tonsky could we get this one reviewed sooner rather than later? It has some fixes that a few users have just pinged me about again

PR #67, opened 2022-08-05
Color highlighter

fbgallet:


panterarocks49 (changes requested):
Did you mean to remove Bionic text in this PR? If you are removing Bionic text please do so in a separate PR

fbgallet:
I'm not removing Bionic text from the main branch, but from this branch. Bionic text is on the main branch, and now I create a new branch for each new extension. But I'm probably doing it wrong, since what is already in the main branch is duplicated unnecessarily.

panterarocks49:
If you remove it from your branch, it will remove it from our branch when you merge it, take a look at the changed files in the PR. I think you just have to update your branch to our main branch, then apply your changes (the new extension). In the github UI I have something to update my fork, not sure if that applies to branches though
image

fbgallet:
Ok, so I updated the branch, but that doesn't recover bionic-text.json. So I copied it in this branch. And I suppose that I need to copy all my .json in all my branches ?

panterarocks49:
yeah, if it's missing it, the "files changed" should only include the current extension being changed

panterarocks49 (changes requested):
https://github.com/fbgallet/roam-extension-color-highlighter/blob/d7917872568829e1968a688f9ca67c4326e2c60c/build.sh#L2
This needs to be `roamjs-scripts build --depot`

panterarocks49:
> Select some part of your text inside a block, add Highight or bold with native Roam hotkeys (Ctrl or Cmd + h for highlight, Ctrl or Cmd + b for bold), then press a letter corresponding to the first letter of the color (without pressing Ctrl-Cmd anymore), e.g. 'g' for green. If you press Shift when pressing the color first letter, the color will be more vivid and the text color in the highlight will be white instead of black.

I can't get this to work, I typed some text, highlighted it, pressed `cmd-h`, let go of held keys, pressed `f`, and it just overwrote my text. Am I missing something?
I can use it through the command palette commands

fbgallet:
Oh, I'm working on Windows, I've not tested on Mac. I suppose that I need to add a test on 'event.metaKey' and not only "event.ctrlKey" !

fbgallet:
I've pushed an update, I hope it will properly listen to 'cmd+h' on Mac.

panterarocks49 (approved):
Awesome, works now, thanks!

PR #74, opened 2022-08-06
Add Unsplash import extension

mlava:


tonsky:
ReferenceError: sendConfigAlert is not defined

mlava:
@tonsky 
Sorry! fixed now, cheers!

tonsky:
So here’s one problem:

Screenshot 2022-08-09 at 15 22 28

It feels like Display mode and Import mode have default values, but in fact they do not. How about you give them default values, and if they are not filled by user, default is used (same for width?)

Another problem: seems like prompt is not supported?

Screenshot 2022-08-09 at 15 24 17

Another problem: if I type gibberish in settings, no validation or error reporting happens

Screenshot 2022-08-09 at 15 25 09
 
Screenshot 2022-08-09 at 15 25 01

Finally, if I set import to random, I get 'undefined' as picture title:

Screenshot 2022-08-09 at 15 24 39

tonsky:
Prompt still doesn’t work for me:

Screenshot 2022-08-10 at 22 56 09

Is it my problem? Does it work for you? If not, let’s remove the setting. Otherwise looks fine, thanks!

mlava:
https://www.loom.com/share/d8dde81117704543aaa1f250ff2db6af

It definitely works for me. I thought prompt was a generic js thing supported by all browsers?? Which OS and browser are you using?

tonsky:
Ha! I thought it’s an error returned from Unsplash API, but it is, in fact, Electron thing https://stackoverflow.com/questions/38858879/replace-prompt-in-on-electron

But since we do have an official electron app, we’ll have to replace it with something

mlava:
@tonsky 
I've updated to use an external prompt script and packaged with webpack.
I think this version should be ok, but keen for feedback.
Cheers!

tonsky:
@mlava new version works great!

Unfortunately, we can't check minified source (one of the main points of the review process is to make sure there are no security holes, and we look at the source code for that).

But we do accept extensions that have build process.

What you need to do is:

- Remove compiled `extension.json` from the repo
- Put all the sources and dependencies relevant to the build into the repo
- Write `build.sh` script that will produce `extension.json` when executed

Thanks!

mlava:
@tonsky 
Thanks, I think the correct files are in the repo now.
Cheers!

tonsky:
Looks good to me—thanks!

PR #77, opened 2022-08-09
Json webhooks

mlava:
@tonsky 
I hope I've addressed all of the feedback from your previous review. Cheers!


tonsky:
You are still concatenating strings here:

```
let q = `[:find (pull ?page [:block/string]) :where [?page :block/uid "${startBlock}"]  ]`;
var thisBlockInfo = await window.roamAlphaAPI.q(q);
```

Since you are only using this query to get block/string, this might work better:

```
window.roamAlphaAPI.data.pull("[:block/string]", [":block/uid", startBlock]);
```

Second concern: default values in config. Maybe you could give delimited a default value?

tonsky:
Thanks! Merged

PR #78, opened 2022-08-09
Add Quick Capture with Todoist extension

mlava:
@tonsky 
I believe I've addressed the issues raised in the previous PR.
Cheers!

### DIFF
--- a/extensions/mlava/quick-capture.json
+++ b/extensions/mlava/quick-capture.json
@@ -1,0 +1,10 @@
+{
+  "name": "Quick Capture with Todoist",
+  "short_description": "Use Todoist as a Quick Capture gateway for Roam Research",
+  "author": "Mark Lavercombe",
+  "tags": ["todoist", "capture", "import"],
+  "source_url": "https://github.com/mlava/quick-capture",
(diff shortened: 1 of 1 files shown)


tonsky:
This

```
let q = `[:find (pull ?page [:block/children]) :where [?page :block/uid "${uid}"]  ]`;
var results = await window.roamAlphaAPI.q(q);
```

Can be replaced with pull:

```
window.roamAlphaAPI.data.pull("[:block/children]", [":block/uid", uid]);
```

Second issue: fetch status should be checked and reported to user if any request fails? It seems like it will fail silently now

mlava:
@tonsky 
Thank you.
I believe I've corrected those issues. I've also handled default config settings better than before, as well as gibberish in settings when needed.

tonsky:
Looks like you forgot `key` argument in

```
function sendConfigAlert() {
```

And I am still getting these errors if I didn’t configure API token

Screenshot 2022-08-10 at 22 59 10

mlava:
@tonsky 
Sorry, fixed now

tonsky:
Seems to work, thanks!

PR #80, opened 2022-08-09
Send to Readwise extension

mlava:
@tonsky 
I believe I've addressed the feedback you gave me on the prior version.
Cheers!


tonsky:
Same notes as for other extensions:

- Replace this `let q = '[:find (pull ?page [:block/string]) :where [?page :block/uid "${startBlock}"]  ]'` with pull
- Give settings default values where it makes sense. Ideally I should only provide access token and not touch anything else
- This part `.catch(error => console.error('error', error));` should provide user-visible error. Otherwise your extension will just fail silently for the user

tonsky:
Seems to work. Thanks!

PR #82, opened 2022-08-10
update source_commit to ptn v0.0.6

scottrblock:
v0.0.5 overhauls how attached images and documents from telegram work. when phonetonote receives a message with an attached file from telegram, i host it on a cloudflare bucket. when syncing the messages in roam, i now use `window.roamAlphaAPI.util.uploadFile` to store the image/file with the rest of the user's roam files. i can then delete the cloudflare image after syncing, which I do.


scottrblock:
closing this to reopen a branch with v0.0.6, which better handles the edge case of unsuccessfully fetching a ptn key as you update your settings

PR #83, opened 2022-08-10
IMDb import

mlava:


8bitgentleman:
This is a cool idea for a plugin and one I would definitely use! Ran into a few things when testing and had some suggestions:

1. Would it be possible to make the metadata format configurable, maybe with some kind of template system? For example in my graph I don't nest metadata, I use `::` Attributes, and I use `Writers` instead of `Writer`. Letting the format be configurable would allow everyone to customize for their own graph :)
2. I ran into silent errors maybe 50% of the time and nothing in the console. Sometimes like the attached image but most of the time I got a "No movie selected!" error even if I was using the exact same movie page and the modal had not popped up
Screen Shot 2022-08-10 at 8 00 10 AM

mlava:
@8bitgentleman 

Thanks for feedback.

Uploading a new version now, with configurable settings for director, writer, cast, year, genre fields and an option to output attributes:

image

8bitgentleman:
Definitely working better! Biggest issue I'm seeing is that attributes still come in with the `**` bolding
image

mlava:
> Definitely working better! Biggest issue I'm seeing is that attributes still come in with the `**` bolding image

fixed in new version

panterarocks49:
@8bitgentleman thanks for the peer review!

panterarocks49:
@mlava I'm getting some errors, any idea why? Looks like the API isn't returning anything

image

Points to this code
image

Maybe there is something wrong with my API key? it's `f5314efa` if you want to try with it

mlava:
That's really weird. I tried with your API key and it works fine:

https://www.loom.com/share/9e785987efb7457cb51afaae59e8ef1c

panterarocks49:
hmm weird, I'll try again in a minute.

btw, you don't need to prefix your settings keys with the extension, internally we separate extension settings. No need to change it, but for future reference you can save some time typing

mlava:
> btw, you don't need to prefix your settings keys with the extension, internally we separate extension settings. No need to change it, but for future reference you can save some time typing

Thanks, that's helpful. Do you also separate extension function names? I assume not as they can be called from anywhere. I have been trying to make sure my function names are different...

panterarocks49:
Interesting, now it works!

mlava:
Very strange! ¯\_(ツ)_/¯

panterarocks49:
Maybe it took some time for imdb to propagate the activation

> Thanks, that's helpful. Do you also separate extension function names? I assume not as they can be called from anywhere. I have been trying to make sure my function names are different...

Internally we import your code as an esm module, so the functions should be separate, at least I can't find the functions from this extension inside of the js console

mlava:
> Internally we import your code as an esm module, so the functions should be separate, at least I can't find the functions from this extension inside of the js console

ok, cool

panterarocks49:
The only way they should be accessible is if you add them onto the window, like smartblocks does

mlava:
> The only way they should be accessible is if you add them onto the window, like smartblocks does

👍

panterarocks49 (approved):

PR #86, opened 2022-08-11
Replace Block Reference with original and bring nested items

mlava:


panterarocks49:
Roam already has this functionality, is it that you want a command palette option for it?

It's currently accessible from here 
image

mlava:
I think I had forgotten that, or I wrote the SmartBlock originally at someone's request and didn't know about the RR inbuilt option.
I don't think it warrants a Command Palette option.
I'll close this PR.
Cheers!

PR #87, opened 2022-08-12
ptn v0.0.7

scottrblock:
v0.0.5 overhauls how attached images and documents from telegram work. when phonetonote receives a message with an attached file from telegram, i host it on a cloudflare bucket. when syncing the messages in roam, i now use `window.roamAlphaAPI.util.uploadFile` to store the image/file with the rest of the user's roam files. i can then delete the cloudflare image after syncing, which I do.

v0.0.6 better handles the edge case of unsuccessfully fetching a ptn key as you update your settings

v0.0.7 adds attachmentUrls to smartblocks

### DIFF
--- a/extensions/phonetonote/phonetonote.json
+++ b/extensions/phonetonote/phonetonote.json
@@ -5,6 +5,6 @@
   "tags": ["phonetonote", "ptn"],
   "source_url": "https://github.com/phonetonote/ptn-roam-depot",
   "source_repo": "https://github.com/phonetonote/ptn-roam-depot.git",
-  "source_commit": "acc1bc932e88876da660c62d9780d0cda3a6f54e",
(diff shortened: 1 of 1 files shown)


panterarocks49:
I get a coors error when I try to upload a file, any ideas?
image

scottrblock:
Ah, I implemented this for telegram images/docs, but it appears to be messing up sms images from twilio (it's trying and failing to upload them). My main development machine is currently being repaired, but I'll take a look at this when i get it back, hopefully in the next few days.

scottrblock:
@panterarocks49 it should be good to go now. sms images (e.g. mms) keep their original twilio url's now. messages from telegram support the more robust image/document handling, where we temporarily upload the file to `https://ptn-telegram-attachments.ptn.computer/`, then upload to roam files

panterarocks49 (approved):
Cool, works for me now

PR #90, opened 2022-08-13
Create yesterday & tomorrow nav buttons

mlava:


pablco:
@mlava It would be cool if there were another two buttons to go to the Next/Previous DNP that exists. What do you think?

mlava:
@pablco
That's interesting. Let me think on it.
It might warrant a separate extension, as this one is intended to create yesterday today and tomorrow buttons that remain fixed rather than changing depending on which page you're on.
Will consider and get back to you.

panterarocks49:
How does this work https://github.com/mlava/yesterday-tomorrow/blob/8912f160d00535cd6ff7a2ee2a79bf18489576b7/extension.js#L163?
I don't see you define that variable anywhere

mlava:
sorry
fixed in e14c6f89e6f7c960b395bae347f97ce22fd54b3c

panterarocks49:
https://github.com/mlava/yesterday-tomorrow/blob/e14c6f89e6f7c960b395bae347f97ce22fd54b3c/extension.js#L137
This will not work right with offline graphs, could you fix to use offline instead of app, you can find the info in `window.roamAlphaAPI.graph.type`
here is an example of using it https://github.com/shodty/roam-breadcrumbs/blob/f500479dfae7ffe56b59e41399508505bc4981c8/extension.js#L119

mlava:
> https://github.com/mlava/yesterday-tomorrow/blob/e14c6f89e6f7c960b395bae347f97ce22fd54b3c/extension.js#L137 This will not work right with offline graphs, could you fix? you can find the info in `window.roamAlphaAPI.graph.type` here is an example of using it https://github.com/shodty/roam-breadcrumbs/blob/f500479dfae7ffe56b59e41399508505bc4981c8/extension.js#L119

thanks for pointing to the fix

I've updated the code and PR

panterarocks49:
no problem, also need to fix here https://github.com/mlava/yesterday-tomorrow/blob/a100ee953ebebc95fdc01de5549d97ad099b29e6/extension.js#L137

mlava:
> no problem, also need to fix here https://github.com/mlava/yesterday-tomorrow/blob/a100ee953ebebc95fdc01de5549d97ad099b29e6/extension.js#L137

Sorry Josh.
Fixed, and checked for other instances and fixed those as well.
Cheers

panterarocks49 (approved):
Thanks!

mlava:
> https://github.com/mlava/yesterday-tomorrow/blob/e14c6f89e6f7c960b395bae347f97ce22fd54b3c/extension.js#L137 This will not work right with offline graphs, could you fix? you can find the info in `window.roamAlphaAPI.graph.type` here is an example of using it https://github.com/shodty/roam-breadcrumbs/blob/f500479dfae7ffe56b59e41399508505bc4981c8/extension.js#L119

Hey Josh. A user has pointed out that the links open in the browser even when clicked in Electron app. Is there a way to stop that happening? 

I found window.roamAlphaAPI.platform.isDesktop but am not sure how to configure the links after recognising the platform... the console suggests the page is like this:
https://roamresearch.com/?server-port=3333#/app/RoamScripts/page/Ed6sa1Zlf
But how do I form a link to another page?

panterarocks49:
@mlava Sorry! I should have pointed you to this API function https://roamresearch.com/#/app/developer-documentation/page/Svn6FFwsW, this will open the links correctly no matter the platform. You just have to attach a click handler and call the function.

mlava:
@panterarocks49 
No worries, I was aware of that function but sometimes you think you'll do something one way and then close your brain to another.
Thanks for pointing me there. PR with fix incoming.
Cheers!

PR #91, opened 2022-08-13
Focus Mode extension

mlava:


panterarocks49:
What do you think about having a single command for toggling focus mode on / off, instead of two separate ones?

mlava:
Like toggle focus mode?

It would need to be able to track toggle state. Where would I track it? In document object maybe?

panterarocks49:
In a local variable would be fine, I'd assume if I refreshed the page it wouldn't save focus mode on.

Up to you if you want to change it, was just thinking it'd be nicer to be able to type "focus" and hit enter to toggle it, as opposed to choosing from two options

mlava:
Update to use a single Command palette option Toggle Focus Mode.
Cheers!

panterarocks49:
awesome, one more thing, `onunload` you should call this function https://github.com/mlava/focus-mode/blob/3cb152714d5f85b25246bcf54bfe157e20efe4d0/extension.js#L104 to remove focus mode, in case it is still enabled.

panterarocks49:
I think you need to refactor the function up a level for that to work, I get a reference error
Also, none of those functions need to be async

image

Thanks!

mlava:
> I think you need to refactor the function up a level for that to work, I get a reference error Also, none of those functions need to be async
> 
> image
> 
> Thanks!

Sorry. Now fixed (I think) in https://github.com/mlava/focus-mode/commit/ab82ee0d93d9c5a449906bfdb6e794a7e14bf51a
Cheers

panterarocks49 (approved):

PR #92, opened 2022-08-14
Elapsed time calculator

fbgallet:


panterarocks49 (changes requested):
I get an error when running "Elapsed Time"

image

I am sure that I have smartblocks installed and the command from the store

Also get an error when trying to click this button (the top one, the bottom one works to insert the first timestamp)
image

image

fbgallet:
And what if you run "Elapsed time" from the command palette ?

fbgallet:
Are you running "Elapsed time" while the cursor is in a block with a timestamp ?

panterarocks49:
Yes in this block
image

fbgallet:
Have you set categories uid ?
I can't reproduce the error.
There is some dependencies imported from .`/utils.js` , is it ok ?

panterarocks49:
> Have you set categories uid ?

No, let me try that

> There is some dependencies imported from ./utils.js , is it ok ?

Yes, it doesn't have to be in a single file before compilation, just can't dynamically import anything

fbgallet:
Ok, i've fixed the small issue !

panterarocks49 (approved):

PR #93, opened 2022-08-14
Daily Calvin & Hobbes

mlava:


panterarocks49 (approved):
https://github.com/mlava/calvin-hobbes/blob/9551190af9a994914d541cead3121e9ace6eb06f/extension.js#L14
this can be undefined sometimes, would recommend choosing a first block on the page if so. I got confused because I wasn't focused on a block
Going to merge anyway, but you should fix this eventually

mlava:
@panterarocks49 
I tried to fix this with window.roamAlphaAPI.ui.mainWindow.getOpenPageOrBlockUid();
Is it expected behaviour that the result is null on a DNP?

PR #96, opened 2022-08-15
Matter: Adds article level note to article page

hunterclarke:
We've recently launched a new feature where a user can add a note at the top level of an article. This is useful for setting the context of the save, general takeaways, etc. This adds note to the article page inside of Roam.


hunterclarke:
Hey @panterarocks49, this is good to go whenever you have a minute.

panterarocks49 (approved):

PR #99, opened 2022-08-18
Add numbered list extension

mlava:


panterarocks49 (approved):
Note that because you aren't using an observer if you navigate away from the page and back the refresh button will not work anymore. Personally I'd just remove the button altogether but up to you. Other than that looks good

PR #100, opened 2022-08-18
New extension: Banner Header Image

mlava:


panterarocks49:
Does anything here look wrong to you? I can't get it to render, I tried navigating away and back. The image on the page is the same url as the one I am using for the banner, so I don't think that is the issue

image

Do you have a test image I could try?

mlava:
I tested with these two:

https://images.unsplash.com/photo-1637305500355-a3f11172cc19?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=1776&q=80

https://images.unsplash.com/photo-1660079543225-b359aeb10988?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=1175&q=80

I suspect that won't be the issue though. Is there anything in console? If you check the page elements can you see the bannerDiv?
Mine looks like:

`<div class="bannerDiv" style="background: url(&quot;https://images.unsplash.com/photo-1660079543225-b359aeb10988?ixlib=rb-1.2.1&amp;ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&amp;auto=format&amp;fit=crop&amp;w=1175&amp;q=80&quot;) center center no-repeat; height: 150px;"></div>`

mlava:
I was actually working on some updates last night, so why don't we pause review of this one while I update and then see if that fixes the issue anyway? @panterarocks49

panterarocks49:
I couldn't find the div on the page, and no errors in the console
> I was actually working on some updates last night, so why don't we pause review of this one while I update and then see if that fixes the issue anyway?

Sounds good

mlava:
@panterarocks49 I think it's ready for review again. Let's hope it works fine this time.

panterarocks49:
ok I found the issue
in `setBanner` you have to sleep for a few ms, it's a race condition
```
function sleep(ms) {
    return new Promise(resolve => setTimeout(resolve, ms));
}
```
and then add `await sleep(50);` to the `setBanner` function

mlava:
awesome, thanks.
Is it on triggering the function? Or later in setBanner??

panterarocks49:
Either I think would work, I did it inside of the `setBanner` function

panterarocks49:
btw would be great if we could find a way for it to take up the whole width, I played around with it for a minute and couldn't figure it out, since if you place the banner higher in the tree then the page scrolls past it

mlava:
> btw would be great if we could find a way for it to take up the whole width, I played around with it for a minute and couldn't figure it out, since if you place the banner higher in the tree then the page scrolls past it

do you mean have a pinned header and if you scroll down the header stays visible?

mlava:
@panterarocks49 
Have pushed the fix for race condition, thanks for identifying it

mlava:
> btw would be great if we could find a way for it to take up the whole width, I played around with it for a minute and couldn't figure it out, since if you place the banner higher in the tree then the page scrolls past it

or do you just mean full width like this:
screenshot 476
and
screenshot 475

i had to move it up above the .rm-article-wrapper div

mlava:
updated to insert bannerDiv at correct level to achieve full width

panterarocks49:
That is what I meant but it causes a weird scroll behavior. it shouldn't stick to the top of the page like that

image

Also it should be `await sleep(50)` not just `sleep(50)`

mlava:
@panterarocks49 
I think the latest version allows for full-width embeds and scrolling works properly. https://github.com/mlava/banner-headers/commit/118be8f67c0647009206f1c260e8732049c9c139
Am interested in your feedback, and grateful for help.
Cheers!

panterarocks49:
This is how it shows up for me now, it goes offscreen to the right

image

maybe just revert to the original idea?

mlava:
Ok, @panterarocks49 
I finally cracked it (css is hard...)
I now have it full width in all layouts, responsive to both sidebars. Works in Electron.
Fingers crossed you have the same experience with https://github.com/mlava/banner-headers/commit/2b664eda8c5eea8971859e5dab06b7b750f3351d

panterarocks49:
It still has the scrolling issue that I mentioned above, I'd expect it to work more like notions banner headers, but if that is ok with you we can merge it.

image

mlava:
@panterarocks49 
Fixed to allow scrolling vs fixed header image via settings.
There is still an issue with the native RR article width setting. So, in Small and Medium the banner won't stretch the full width when it is set to scrolling mode. In fixed mode it will stretch full width.
I think this is the best I can do, and we should publish and see what users think and if the Small/Medium scrolling use case bothers people.
Cheers!

panterarocks49:
Great thanks! sorry for the dead end suggestions on this one

mlava:
Hey, no problem @panterarocks49 
I got to see what's new in CSS in the twenty years since I first played with it 👍 
Thanks for approve and merge.

PR #102, opened 2022-08-18
Query Builder: Version 2022-08-18-19-04

dvargas92495:


panterarocks49:
Note that a lot of the queries you changed to fast.q in components would be faster as pulls (assuming you changed for speed reasons)

panterarocks49 (approved):

PR #107, opened 2022-08-19
Publishing ccc-roam-ocr extension

c3founder:


panterarocks49:
Review for [ccc/ccc-roam-language](https://github.com/c3founder/ccc-roam-language)

Are you sure you want the background of the textarea to be yellow? I was very surprised when the extension turned the textarea background color to be yellow. Feels like a separate extension / css snippet

You need to remove the interval in `onunload`, you can just copy paste this code below to fix it
```
let interval;

function onload() {
    interval = setInterval(activateAutoDir, 1000);
}

function onunload() {
    clearInterval(interval);
}

export default {
    onload: onload,
    onunload: onunload
}

```

panterarocks49:
There is something wrong with your stripe account, are you sure you signed up through Roam? and if so make sure you have filled out all the info
It should be the stripe account from this page in Roam
image

panterarocks49:
Review for [ccc/ccc-roam-ocr](https://github.com/c3founder/ccc-roam-ocr)

I went ahead and fixed most of the issues for you in a PR https://github.com/c3founder/ccc-roam-ocr/pull/1, please take a look at the PR, I think everything works except I'm not sure about the key command.

panterarocks49:
More questions on [ccc/ccc-roam-ocr](https://github.com/c3founder/ccc-roam-ocr)

If teserract.js works completely offline, why do you have to rehost the image on your server? Is it because of the math extraction?

panterarocks49:
Review for [ccc/ccc-roam-pdf-1](https://github.com/c3founder/ccc-roam-pdf-1)

PR for fixing pdf https://github.com/c3founder/ccc-roam-pdf-1, it fixes issues with dynamically loading and cleaning up state, please check that it still works correct for you

panterarocks49:
More on [ccc/ccc-roam-pdf-1](https://github.com/c3founder/ccc-roam-pdf-1)

Just to be clear, this doesn't _save_ the highlights on your server right? it saves them in Roam and then sends them to your server for rendering the PDF

How will this interact with v2?
will users have to redo their highlights?
Will it be a smooth upgrade?

c3founder:
> More questions on [ccc/ccc-roam-ocr](https://github.com/c3founder/ccc-roam-ocr)
> 
> If teserract.js works completely offline, why do you have to rehost the image on your server? Is it because of the math extraction?


If I remember correctly when I wanted to send external image link to mathpix I was getting CORS error.

c3founder:
> More on [ccc/ccc-roam-pdf-1](https://github.com/c3founder/ccc-roam-pdf-1)
> 
> Just to be clear, this doesn't _save_ the highlights on your server right? it saves them in Roam and then sends them to your server for rendering the PDF
> 
> How will this interact with v2? will users have to redo their highlights? Will it be a smooth upgrade?

Right, in both versions, roam act as the annotation storage. Nothing stays on the server (OK technically v1 saves images resulted from area highlight on our firebase; I need to update that since the roam upload api is now available) 

Unfortunately, v2 is not backward compatible. v1 uses react-pdf-highlighter which has a non-standard way of storing highlights. I'm using PDFTron viewer for v2 which is more advanced and stores highlights in standard xfdf format and I have change how things are stored in roam for various reason. 

So in short: eventually, I need to write a highlight data converter.

c3founder:
> There is something wrong with your stripe account, are you sure you signed up through Roam? and if so make sure you have filled out all the info It should be the stripe account from this page in Roam image

I had something through BuyMeCoffee, didn't know it won't work. Will fix it soon.

c3founder:
> Review for [ccc/ccc-roam-language](https://github.com/c3founder/ccc-roam-language)
> 
> Are you sure you want the background of the textarea to be yellow? I was very surprised when the extension turned the textarea background color to be yellow. Feels like a separate extension / css snippet
> 
> You need to remove the interval in `onunload`, you can just copy paste this code below to fix it
> 
> ```
> let interval;
> 
> function onload() {
>     interval = setInterval(activateAutoDir, 1000);
> }
> 
> function onunload() {
>     clearInterval(interval);
> }
> 
> export default {
>     onload: onload,
>     onunload: onunload
> }
> ```

Users could change the color (or disable it) using the corresponding CSS. 

Now on roam depot what do you suggest to do instead? Especially for this extension, there are many CSS variables to customize which may be too much for the config panel.

c3founder:
> > There is something wrong with your stripe account, are you sure you signed up through Roam? and if so make sure you have filled out all the info It should be the stripe account from this page in Roam image
> 
> I had something through BuyMeCoffee, didn't know it won't work. Will fix it soon.

I think I fixed it. Lmk if my stripe account is not working. Thanks!

panterarocks49:
> Unfortunately, v2 is not backward compatible. v1 uses react-pdf-highlighter which has a non-standard way of storing highlights. I'm using PDFTron viewer for v2 which is more advanced and stores highlights in standard xfdf format and I have change how things are stored in roam for various reason.

oh that's unfortunate, but fine as long as there is a plan for them users to be able to run both (so they can see the old annotations) or migrate

panterarocks49:
> Now on roam depot what do you suggest to do instead? Especially for this extension, there are many CSS variables to customize which may be too much for the config panel.

Config panel is currently the best option

panterarocks49:
Hey so you still need to update the commit hash inside of this PR for each of the repositories, I think after you update those we are good to go

panterarocks49:
> Now on roam depot what do you suggest to do instead? Especially for this extension, there are many CSS variables to customize which may be too much for the config panel.

Next best option would be to mention what you can customize in the readme and how, maybe a code block with an example

c3founder:
> Hey so you still need to update the commit hash inside of this PR for each of the repositories, I think after you update those we are good to go

OK, I don't know what I need to know, sorry. Can you please be more specific?

c3founder:
I updated the commit hashes. 

Also, submitted the new version of the highlighter.

panterarocks49 (approved):

PR #108, opened 2022-08-19
Universal Selector

fbgallet:
Didn't appear in pull request list anymore, I suppose I've made a mistake in pushing last update this night! But it was pushed here 14 days ago.

### DIFF
--- a/extensions/fbgallet/roam-extension-universal-dropdown.json
+++ b/extensions/fbgallet/roam-extension-universal-dropdown.json
@@ -5,6 +5,6 @@
   "tags": ["list", "dropdown", "database"],
   "source_url": "https://github.com/fbgallet/roam-extension-universal-dropdown",
   "source_repo": "https://github.com/fbgallet/roam-extension-universal-dropdown.git",
-  "source_commit": "da33c90437043693c88ceecc32ae0abd255bd666",
+  "source_commit": "2d6c25638259a58c84b80771b3f521afb2083d05",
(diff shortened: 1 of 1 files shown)


panterarocks49:
I think the PRs auto close if the files changed = 0, kind of annoying but not sure how to disable that

panterarocks49:
I'm not sure what is going wrong but the extension.js is not being output to the top level directory, it is output inside of the `build` folder. 
Since you're not actually using any external deps I'd recommend just copy pasting index.js to the top level and renaming to extension.js

panterarocks49 (changes requested):
In addition to the comment above, I was not able to get this to work, I started by trying the buttons, and then tried the command palette command

image

Here are the errors I get
image

I have smartblocks installed

image

Any ideas?

fbgallet:
You need to install the corrsponding SmartBlock from the Sb store

panterarocks49:
I think you have to remove build.sh as well, the build fails if not

panterarocks49:
> You need to install the corrsponding SmartBlock from the Sb store

Oh right sorry, did not read the README in full

Is there a reason to split it up like that? can you not register the smartblock from the code easily?

fbgallet:
I can register Smartblocks commands but not workflows. By the way, some users were already using the Smartblock and are waiting for the update since Smartblock extension rules changes, and I will release a version (in one or two weeks) that will works without Sb.

panterarocks49 (approved):

PR #109, opened 2022-08-19
Fix YouTube links extension

mlava:


panterarocks49:
Hey what are some examples of broken youtube links? if it's just missing `www` like in your example then we can fix that on our end so you don't have to add it (we already do a bunch of processing on yt links)

mlava:
I get a lot of imports from Readwise that are missing the www. They come
from annotate.tv which I think sends them to Readwise without the www
I think that's my main source of irritation :-D

panterarocks49:
Ok, I'm going to try to fix this in our code and hold off on this extension

mlava:
No worries

panterarocks49:
Released a fix for this today

mlava:
Awesome, thanks!

PR #111, opened 2022-08-19
Copy Codeblock Button

8bitgentleman:
Adds a button to all codeblocks to let you copy their contents to the clipboard


8bitgentleman:
Anyone know why this build is failing? Says there's an illegal white space but doesn't say _where_ it is

extensions/8bitgentleman/copy-code-blocks.json
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "Code Block Copy Button",
+    "short_description": "Adds a button in code blocks to copy their contents to the clipboard.",
+    "author": "Matt Vogel",
+    "tags": ["code block"],
+    "source_url": "https://github.com/8bitgentleman/roam-depot-copy-code-block",
+    "source_repo": "https://github.com/8bitgentleman/roam-depot-copy-code-block.git",
+    "source_commit": "8acd0b22e367dd36ab1909e1911ee37da7add906",
+    "stripe_account": "acct_1LJFEYQmxalymEZL"
+
```
panterarocks49, line 10: Our json parser is not very forgiving, I think this line is the issue

8bitgentleman: I deleted the empty line there but no change change, could it be something else?

panterarocks49: Empty line is still there in files changed 
image

8bitgentleman: ended up being a rogue `.DS_Store` file I think

panterarocks49 (approved):
Small bug, if you want to fix in another PR
If I edit the code block and then click copy, it doesn't copy the current text of the code block
This is because we only update our database on blur

Would love to have this for inline code as well!

PR #114, opened 2022-08-19
Export Tables

8bitgentleman:
Adds a button to export a table to CSV. 

Right now the button lives in the first cell of the table. 
Screen Shot 2022-08-19 at 5 50 11 PM
I'd prefer if it could live next to the hover edit button on the far right but every time I added it to the `hoverparent` with `.hoveronly` clicking on the download button would just edit the block text. 
image

If there's a way to do this so editing the block isn't triggered on button click please let me know!


panterarocks49:
> If there's a way to do this so editing the block isn't triggered on button click please let me know!

Add the class `dont-focus-block` to the button

8bitgentleman:
Thanks @panterarocks49 , updated per your suggestion and added support for exporting attr tables

panterarocks49 (approved):

PR #119, opened 2022-08-22
Add LaTeX exporter

alixlahuec:


panterarocks49:
Did you mean to make this a work in progress?

panterarocks49:
do you have plans to migrate zotero-roam as well?

alixlahuec:
> Did you mean to make this a work in progress?

Yes! Wanted to setup everything for submission, but haven't decided if I want to submit it yet

alixlahuec:
> do you have plans to migrate zotero-roam as well?

Also yes, I'm working on making it compatible with Roam Depot now -- much bigger project though so it's taking some time

panterarocks49:
> Also yes, I'm working on making it compatible with Roam Depot now -- much bigger project though so it's taking some time

Awesome, let me know if there is anything I can do to help

panterarocks49 (changes requested):
@alixlahuec Thanks for submitting!

I only see one small thing that you need to fix, which is to remove the block context menu for exporting when your extension unloads here https://github.com/alixlahuec/latex-roam/blob/59610b0f776f7f9023c0fbfe7b53fc11e6f0fbbd/src/components/App.jsx#L23

It would also be nice if you could not commit extension.js and extension.css at the root of the repo, since those are compiled by `build.sh` (not necessary if you are using them but helps reviewing changes to the repo)

I also have a couple of questions
In your wiki, it says I can export images to latex, but it doesn't seem to be working for me, does this not work with images hosted on Roam? 

Here are two examples I tried
```
![](https://firebasestorage.googleapis.com/v0/b/firescript-577a2.appspot.com/o/imgs%2Fapp%2Ftest103%2FZkwm2ZDBmZ.png?alt=media&token=21daed0b-1acc-4935-a42b-1eb7cf55f2c6)

![](https://images.unsplash.com/photo-1611915387288-fd8d2f5f928b?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxleHBsb3JlLWZlZWR8MXx8fGVufDB8fHx8&w=1000&q=80)
```

If I added a page context menu API (similar to the block context menu API), so that you could add the button to this menu

image

Would you move the button into that menu? Asking because the latex button creates a lot of visual noise that I imagine would be annoying to have on at all times

alixlahuec:
> @alixlahuec Thanks for submitting!
> 
> I only see one small thing that you need to fix, which is to remove the block context menu for exporting when your extension unloads here https://github.com/alixlahuec/latex-roam/blob/59610b0f776f7f9023c0fbfe7b53fc11e6f0fbbd/src/components/App.jsx#L23
> 
> It would also be nice if you could not commit extension.js and extension.css at the root of the repo, since those are compiled by `build.sh` (not necessary if you are using them but helps reviewing changes to the repo)
> 
> I also have a couple of questions In your wiki, it says I can export images to latex, but it doesn't seem to be working for me, does this not work with images hosted on Roam?
> 
> Here are two examples I tried
> 
> ```
> 
> 
> 
> ```
> 
> If I added a page context menu API (similar to the block context menu API), so that you could add the button to this menu
> 
> image
> 
> Would you move the button into that menu? Asking because the latex button creates a lot of visual noise that I imagine would be annoying to have on at all times

Thanks for catching, I'll fix that!
For the `build.sh` / `extension.*` files, I use the `extension.css` / `extension.js` for local testing & development so I'd prefer keeping them if possible (so that I don't have to run the build step every time I've made some changes and want to test), but yeah if it's a bother I can remove them.

For your questions:
- The plugin should be able to get images hosted in Roam, because that's the example I used to build the functionality. I haven't done extensive testing though so might be a bug, I'll look into it 👍 
- Re: page context menu - yep, that would solve that for me! I would do the same thing as for the block context menu

alixlahuec:
Added a cleanup step to remove the block context menu option when the extension unmounts.

I also looked into the issue with images -- found & fixed a bad regex that was mangling images without description (`` as opposed to `some description`) and preventing them from being processed as images.
For the Unsplash link, I figured out what the problem is -- the link doesn't have a file extension in it, so the plugin fails at the matching step. I'll add this to my list of stuff to look into

panterarocks49:
> For the build.sh / extension.* files, I use the extension.css / extension.js for local testing & development so I'd prefer keeping them if possible (so that I don't have to run the build step every time I've made some changes and want to test), but yeah if it's a bother I can remove them.

Yeah sure feel free to keep them

panterarocks49:
> Added a cleanup step to remove the block context menu option when the extension unmounts.

I also looked into the issue with images -- found & fixed a bad regex that was mangling images without description ( as opposed to some description) and preventing them from being processed as images.
For the Unsplash link, I figured out what the problem is -- the link doesn't have a file extension in it, so the plugin fails at the matching step. I'll add this to my list of stuff to look into

Ok great this looks like it is good to go, the Roam image is now working for me

panterarocks49 (approved):

PR #120, opened 2022-08-22
Create Shortcode Embed extension

mlava:


panterarocks49 (changes requested):
Could you add something to the readme telling people they can use it via the command palette? I had to look at the code to figure that out

Other than that looks good

panterarocks49:
Maybe also change it to say "Paste Embed from clipboard", instead of just "Embed from clipboard"

mlava:
Thanks @panterarocks49 
Both changes made and pushed to https://github.com/mlava/shortcode-embeds/commit/920a1784afba57857f6ffe8586180b65517e8e6a

panterarocks49 (approved):
Thanks!

PR #121, opened 2022-08-22
Fixed temperatures displaying incorrectly and CSS in Weathercards

mlava:


mlava:
@panterarocks49 
This version fixes the issue you had with temperatures being way too high...

panterarocks49 (approved):

PR #123, opened 2022-08-22
Memo - Spaced Repetition for Roam

digitalmaster:
CleanShot 2022-08-22 at 19 33 22@2x

Had 2 weeks off from work and decided to build the plugin I always wished Roam had built in. A simple, easy-to-use, clean, spaced repetition UI. There's still a lot I want to add but I think this is at a ok place to publish and iterate from there.

Looking forward to some feedback 🙏🏽

### DIFF
--- a/extensions/digitalmaster/roam-memo.json
+++ b/extensions/digitalmaster/roam-memo.json
@@ -5,6 +5,6 @@
     "tags": ["spaced repetition"],
     "source_url": "https://github.com/digitalmaster/roam-memo",
     "source_repo": "https://github.com/digitalmaster/roam-memo.git",
-    "source_commit": "07a6ded9cbd2065ac4f515ec1fc1ab9bbac276b8",
+    "source_commit": "f851a9ffbe350596866f560c4e9b74a82a37f9c5",
     "stripe_account": "acct_1LZNprQe7Pm7RMMV"
   }
--- a/extensions/digitalmaster/roam-memo.json
+++ b/extensions/digitalmaster/roam-memo.json
@@ -5,6 +5,6 @@
     "tags": ["spaced repetition"],
     "source_url": "https://github.com/digitalmaster/roam-memo",
     "source_repo": "https://github.com/digitalmaster/roam-memo.git",
-    "source_commit": "77b752059f75a777f419aab4fe7c134cc600d599",
+    "source_commit": "07a6ded9cbd2065ac4f515ec1fc1ab9bbac276b8",
     "stripe_account": "acct_1LZNprQe7Pm7RMMV"
   }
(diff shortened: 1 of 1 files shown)


panterarocks49:
Hey thanks for submitting to Roam Depot!

I am just starting your review and noticed this
https://github.com/digitalmaster/roam-memo/blob/cbb3c484f170af024400f33439d81853c34fb35f/src/extension.js#L21
well you can do that you should store it under a name unique to your extension, something like `window.roamMemo.extensionAPI`, so that it doesn't conflict with other extensions

digitalmaster:
> Hey thanks for submitting to Roam Depot!

Hey @panterarocks49 👋🏽 - Thanks for taking the time to review.

> I am just starting your review and noticed this https://github.com/digitalmaster/roam-memo/blob/cbb3c484f170af024400f33439d81853c34fb35f/src/extension.js#L21 well you can do that you should store it under a name unique to your extension, something like `window.roamMemo.extensionAPI`, so that it doesn't conflict with other extensions

Ah! Great call... Will Update 🙏🏽

panterarocks49:
Is there a reason these are being tagged with [[date}]] instead of just [[date]]? Looks like a typo maybe

image

panterarocks49:
Couple of more comments (non blocking but thought I'd point out)

https://github.com/digitalmaster/roam-memo/blob/cbb3c484f170af024400f33439d81853c34fb35f/src/queries.js#L172
You should be able to `await` here instead of sleeping and waiting for the block to be created (same goes for any write event from Roam)

https://github.com/digitalmaster/roam-memo/blob/cbb3c484f170af024400f33439d81853c34fb35f/src/queries.js#L16
Not sure this matters but `q` is sync and you shouldn't need the `await` here (same goes for any read event from Roam)

https://github.com/digitalmaster/roam-memo/blob/cbb3c484f170af024400f33439d81853c34fb35f/src/queries.js#L322
although createBlock does not return the uid of the block created you can create your own uid and pass it in, then return that from your wrapper function
Something like this
```
const uid = window.roamAlphaAPI.util.generateUID();
await window.roamAlphaAPI.createBlock({
    location: { 'parent-uid': parent_uid, order: order },
    block: { string: block, uid: uid, ...blockProps },
});

```

panterarocks49 (changes requested):
Alright, finished my review, overall love this extension! Super clean and easy to use

Only thing you have to change / address is my first two comments above, you can ignore or fix the third comment later

digitalmaster:
> Is there a reason these are being tagged with [[date}]] instead of just [[date]]? Looks like a typo maybe
> 
> image

Nice Catch.. definitely a typo.

digitalmaster:
Excellent feedback! Really appreciate you taking a look. I've pushed changes (thanks for the tip on creating hash manually.... will refactor that one next week along with some other changes). 🙌🏽🙏🏽

panterarocks49 (approved):
Sweet, looks good now

PR #126, opened 2022-08-24
New extension: Pexel Image embed

mlava:


panterarocks49:
Could you not use @omiu/dialog-extention, requiring this loads all of omiu, which seems to be a competing framework to react.
We provide blueprint which has an equivalent component to the one you are using https://blueprintjs.com/docs/versions/3/#core/components/alert

mlava:
I can't see a prompt equivalent in the blueprints docs. Do you know if there is one? Alert won't be enough... I need the a prompt and electron doesn't allow native js prompts

mlava:
> I can't see a prompt equivalent in the blueprints docs. Do you know if there is one? Alert won't be enough... I need the a prompt and electron doesn't allow native js prompts

I thought I was just importing '@omiu/dialog-extention'; ?

In Line 1 of src/index.js that is the import

panterarocks49:
> > I can't see a prompt equivalent in the blueprints docs. Do you know if there is one? Alert won't be enough... I need the a prompt and electron doesn't allow native js prompts
> 
> I thought I was just importing '@omiu/dialog-extention'; ?
> 
> In Line 1 of src/index.js that is the import

Well yeah but it brings 5 dependencies along with it, see npm for the dependencies https://www.npmjs.com/package/@omiu/dialog-extention

mlava:
Ok, back to the drawing board

panterarocks49:
> I can't see a prompt equivalent in the blueprints docs. Do you know if there is one? Alert won't be enough... I need the a prompt and electron doesn't allow native js prompts

I think you can make a prompt with an alert by just putting an input inside of it, but if you don't want to do that some people use https://github.com/marcelodolza/iziToast which is a better option, I believe it supports an input inside of a popup (although it's not directly called a prompt)

mlava:
Ok, cool, thanks

panterarocks49:
I'm not too familiar with izToast but this code has some examples https://github.com/fbgallet/roam-extension-find-replace/blob/main/src/index.js#L86

and this example might help with the alert https://github.com/palantir/blueprint/blob/develop/packages/docs-app/src/examples/core-examples/alertExample.tsx#L76, basically you'd just replace the `p` with and `input` and control the state, but I've never done that before so not sure it will look good

mlava:
Please ignore this extension for now. Thinking I might merge with Unsplash instead of having separate. Cheers!

PR #128, opened 2022-08-25
New extension: Workspaces

mlava:


panterarocks49:
You're missing `build.sh` which should `npm install` and `npm run build` inside of it

I think you might be missing some css?
image

Some general feedback, but not blocking

I would expect that clicking on the workspace button, even if it is already selected would bring me back to that workspace, currently I'd have to go to another workspace and then back

I might be nice to store the data as a reference instead of just the uid, then I could easily see what the workspace is made up of

Overall, love this concept!

mlava:
Sorry, @panterarocks49, that was a noob mistake.
Required files are now included.
I think that's a good idea on the blockref... will work on that. I've been working on some updates and am aware of the selection menu issue and will figure out something.
Have also implemented kb shortcuts in new version which might make the selector issue less irritating.
Cheers!

mlava:
> You're missing `build.sh` which should `npm install` and `npm run build` inside of it
> 
> I think you might be missing some css? image
> 
> Some general feedback, but not blocking
> 
> I would expect that clicking on the workspace button, even if it is already selected would bring me back to that workspace, currently I'd have to go to another workspace and then back
> 
> I might be nice to store the data as a reference instead of just the uid, then I could easily see what the workspace is made up of
> 
> Overall, love this concept!

@panterarocks49 
I think I've got this pretty ready to go now. I've tested and re-tested, ironed out some bugs etc.
With regard to the storing uid as reference, do you just mean as ((uid)). That shows something for blocks but not pages for me. I wasn't sure if you meant something else?

panterarocks49:
> With regard to the storing uid as reference, do you just mean as ((uid)). That shows something for blocks but not pages for me. I wasn't sure if you meant something else?

My thinking was either store the [[page title]] for pages or the ((uid)) for blocks, but I realize that's kind of a lot of code with possibly not a lot of benefit

panterarocks49:
I'm still getting the css bug, I think you might have to import the css via css-loader https://webpack.js.org/loaders/css-loader/
and then import like this https://github.com/fbgallet/roam-extension-find-replace/blob/main/src/index.js#L2

You could also copy paste the css into your extension.css file, that might be easier than setting up css-loader

image

mlava:
sorry @panterarocks49 
fixed

panterarocks49 (approved):
Great, thanks, looks good now

PR #129, opened 2022-08-25
Update show-favicon with the correct stripe account

paulovieira:
It seems I messed up something before when trying to create the express stripe account. I probably didn't choose the country correctly.

Anyway, I managed to created / finish the stripe account using a bank account in Portugal, with no problems.

More details here: https://github.com/Roam-Research/roam-depot/pull/48#issuecomment-1214469738


panterarocks49:
Stripe account looks good now, transferred the funds from last month

paulovieira:
> Stripe account looks good now, transferred the funds from last month

That was very kind. Thanks!

PR #139, opened 2022-08-27
Roam Studio

rcvd:
Initial Release

### DIFF
--- a/extensions/rcvd/RoamStudio.json
+++ b/extensions/rcvd/RoamStudio.json
@@ -4,6 +4,6 @@
   "author": "Alexander Rink",
   "source_url": "https://github.com/rcvd/RoamStudio",
   "source_repo": "https://github.com/rcvd/RoamStudio.git",
-  "source_commit": "11f8e8ee6ec3b9030d977c24a49b7c1ea6dfc989",
+  "source_commit": "ddbad3bd6fdf85e2a587f917d5483e72759ce59d",
   "stripe_account": "acct_1LPgSxQUtJ9VpaGT"
 }
--- a/extensions/rcvd/RoamStudio.json
+++ b/extensions/rcvd/RoamStudio.json
@@ -4,6 +4,6 @@
   "author": "Alexander Rink",
   "source_url": "https://github.com/rcvd/RoamStudio",
   "source_repo": "https://github.com/rcvd/RoamStudio.git",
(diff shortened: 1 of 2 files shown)


rcvd:
@panterarocks49: This is the initial release with just one theme. When I add more themes (and fonts), I will quickly exceed the current set size limit of around 4 MB. What would be the recommended workaround for this?

rcvd:
Just added some small bug fixes.

panterarocks49:
> When I add more themes (and fonts), I will quickly exceed the current set size limit of around 4 MB

Who said there was a 4MB limit?

Can you commit the files unminified? Kind of hard to review this way

image

rcvd:
> 


> Who said there was a 4MB limit?
The build process failed. 
Screenshot 2022-08-31 at 06 39 00

> Can you commit the files unminified? Kind of hard to review this way
Updated it with the minified files.

panterarocks49:
Interesting, that is a size limit enforced by roamjs-scripts, roam depot does not have a size limit. I asked David if there is a way to override that.

panterarocks49:
I submitted a PR to your repo fixing the size issue https://github.com/rcvd/RoamStudio/pull/1, you should be able to add the other themes now

panterarocks49:
Aside from the above PR, the only other thing that I noticed that needs to be changed is the dom IDs, they should be prefixed by a name unique to the extension.

https://github.com/rcvd/RoamStudio/blob/master/src/build.js#L410

So instead of `"css-theme"` it should be something like `"roamstudio-css-theme"`, to avoid conflicts with other extensions

rcvd:
Merged PR and updated IDs.

panterarocks49:
Thanks! 

I noticed you changed all ' to " and that had unintended affects on icons-feather, it looks like it broke some of the svgs
https://github.com/rcvd/RoamStudio/blob/master/src/RoamStudio/modules/icons-feather.css#L85

Although it's not used in the PR, so this is still good to merge

panterarocks49 (approved):

panterarocks49:
@rcvd Shoot, I missed this in the re-review, I am getting an error on unloading RoamStudio because some of the IDs were not updated, here https://github.com/rcvd/RoamStudio/blob/81ca3433521b8f9f94623030070616bc7cc7a6fa/src/build.js#L420
Could you update those IDs sooner than later in a new PR?

rcvd:
Will do, thanks for input.

PR #141, opened 2022-08-28
Custom Mobile Buttons

8bitgentleman:
An extension to add custom buttons to the bottom bar on mobile. First time using a reactComponent in the settings panel, not sure about the styling.
Also if there is a simple way to toggle between formatting (eg bold on/off) I haven't figured it out

### DIFF
--- a/extensions/8bitgentleman/mobile-buttons.json
+++ b/extensions/8bitgentleman/mobile-buttons.json
@@ -1,0 +1,10 @@
+{
+    "name": "Custom Mobile Buttons",
+    "short_description": "Adds custom buttons to the bottom bar of Roam on mobile",
+    "author": "Matt Vogel",
+    "tags": ["mobile", "smartblocks"],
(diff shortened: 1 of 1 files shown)


panterarocks49:
> Also if there is a simple way to toggle between formatting (eg bold on/off) I haven't figured it out

There isn't

panterarocks49:
Small thing I noticed, sometimes I get [object object] displayed here, but doesn't seem to break anything

image

panterarocks49:
Other than that, it looks good to me, lmk if you want me to merge it with that bug or fix it

8bitgentleman:
Thanks for the review! I have an idea for a better solution that may get rid of [object object], I'll look into it this week

panterarocks49:
Do you want me to merge it in the meantime? seems like a small enough bug that it won't effect users

8bitgentleman:
Sure you can go ahead and merge

PR #147, opened 2022-08-29
Update: Fix failure to close task bug in Todoist Task Management

mlava:


panterarocks49:
Changing the id of the settings will reset all of the users settings (they will have to put their API key in again), was that intentional?

mlava:
> Changing the id of the settings will reset all of the users settings (they will have to put their API key in again), was that intentional?

fixed, sorry

panterarocks49 (approved):

PR #149, opened 2022-08-29
Add Roam Date UX plugin

Stvad:
via https://github.com/Stvad/roam-date

### DIFF
--- a/extensions/Stvad/roam-date.json
+++ b/extensions/Stvad/roam-date.json
@@ -12,5 +12,5 @@
     "source_url": "https://github.com/Stvad/roam-date",
     "source_repo": "https://github.com/Stvad/roam-date.git",
     "source_commit": "e6fe27525aa5d3dad7398464f62f5f9c1c98efc4",
-    "stripe_account": "acct_1LcFfTHIh2HKWpPF"
+    "stripe_account": "acct_1LcbYoQmU9n2q3gV"
 }
--- a/extensions/Stvad/roam-date.json
+++ b/extensions/Stvad/roam-date.json
@@ -1,6 +1,6 @@
 {
     "name": "Date UX improvements",
-    "short_description": "Introduces UX improvements for interacting with Dates. Including a widget to easily change dates on mobile, natural language date creation and easier daily notes navigation",
+    "short_description": "Introduces UX improvements for interacting with Dates. Including a widget to easily change dates on mobile, natural language date creation and easier daily notes page navigation",
     "author": "Vlad Sitalo",
     "tags": [
         "date",
--- a/extensions/Stvad/roam-date.json
+++ b/extensions/Stvad/roam-date.json
@@ -11,6 +11,6 @@
     ],
     "source_url": "https://github.com/Stvad/roam-date",
     "source_repo": "https://github.com/Stvad/roam-date.git",
-    "source_commit": "a35edc8008ce49860c5c668f8c4bc4e0c390ab04",
+    "source_commit": "909c40afa256a9144112cf9263da8fdef67d6817",
     "stripe_account": "acct_1LcbYoQmU9n2q3gV"
 }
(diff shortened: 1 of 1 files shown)


panterarocks49 (changes requested):
I can't find the npm package / github repository for `roam-api-wrappers` could you point me to both of those? I need to review that repository as well

There are a couple of different places you need to remove event listeners / hotkeys in `onunload`
https://github.com/Stvad/roam-date/blob/e6fe27525aa5d3dad7398464f62f5f9c1c98efc4/src/navigation.ts#L7
https://github.com/Stvad/roam-date/blob/e6fe27525aa5d3dad7398464f62f5f9c1c98efc4/src/index.ts#L38

The extension.js file size is a lot larger than I'd expect it to be, I think this is because roamjs-components is not being tree shaken correctly due to this import
https://github.com/Stvad/roam-date/blob/e6fe27525aa5d3dad7398464f62f5f9c1c98efc4/src/index.ts#L7
These need to be all separate imports
```
import createBlockObserver from "roamjs-components/dom/createBlockObserver";
...etc
```

Is this config observer used for anything?
https://github.com/Stvad/roam-date/blob/e6fe27525aa5d3dad7398464f62f5f9c1c98efc4/src/index.ts#L40

Your assets in the README should use the full github url, Roam doesn't recognize relative paths

panterarocks49:
Also onunload should remove these calendar buttons
https://github.com/Stvad/roam-date/blob/e6fe27525aa5d3dad7398464f62f5f9c1c98efc4/src/index.ts#L53

Stvad:
Thanks for taking a look!

> Roam API Wrappers

It's https://github.com/Stvad/roam-api-wrappers and package with the same name on npmjs

> removing event listeners, buttons/etc

Is that a hard requirement? I understand that it's nice to do so, but I never bothered to clean up such state in extensions to date, as one can always reload the page.

>The extension.js file size is a lot larger than I'd expect it to be, I think this is because roamjs-components is not being tree shaken correctly due to this import
https://github.com/Stvad/roam-date/blob/e6fe27525aa5d3dad7398464f62f5f9c1c98efc4/src/index.ts#L7
These need to be all separate imports

will check

>Is this config observer used for anything? 

not atm, it's there bc it was part of the template 😛 

> Your assets in the README should use the full github url, Roam doesn't recognize relative paths

👍

panterarocks49:
> Is that a hard requirement? I understand that it's nice to do so, but I never bothered to clean up such state in extensions to date, as one can always reload the page.

Yes that is a hard requirement, both so users can uninstall extensions without reloading the page and so when you update you extension, it updates without reloading the page (and without running two versions at once)

Stvad:
Updated to address the comments :)

panterarocks49:
Thanks! Everything seems good except I still am seeing a size issue, although it is less than before.

See the output of `npx roamjs-scripts build --depot --analyze`
image

lodash is taking up half of the module size (500kb), which is interesting since https://bundlephobia.com/package/lodash@4.17.21 says it should only be ~70kb

I think this is because of this line https://github.com/Stvad/roam-api-wrappers/blob/master/src/data/index.ts#L4 which imports all of lodash
Could you try importing only the functions you need?
I think something like this will work
`import { countBy } from 'lodash';`
but you might need to import from the sublib
`import countBy from 'lodash/collection';`

Stvad:
Will take a look, but those numbers don't make sense to me 🤔. The whole thing built without analyze is ~290kb. buth with analyze it grows to 1.2mb

panterarocks49:
Oh yeah, interesting, didn't realize the extension.js is actually 290kb. No worries if you can't figure it out, 290kb is much better than 1mb!

Stvad:
switching to `lodash-es` helps to get it to 226kb

panterarocks49:
eh, that's not a whole lot, up to you, I read lodash-es has issues

Stvad:
pushed un update for now, may switch back later if encounter those 🤷‍♂️

PR #150, opened 2022-08-29
New extension: Unread Badge

mlava:


panterarocks49 (changes requested):
https://github.com/mlava/unread-badge/blob/ba11d1efe6180125fb09bd28a6f3f22a1cdf6806/extension.js#L165
Need to clear this interval onunload, also store it in a local var instead of the window object to avoid conflicts with other extensions

panterarocks49:
How do I process an unread?

panterarocks49:
Installing was a bit of a weird process, I first got an alert to add a tag, then I added the tag and nothing happened. I think it should be reactive to the settings, I had to reload the page to get it to show the tag

mlava:
Thanks Josh. Will get back to finessing this one a bit more. Cheers!

panterarocks49:
> PS @panterarocks49 Does the API allow set/unset of starred/shortcut status for a page? I can't seem to find it in docs...

I thought it did, but it looks like it doesn't

mlava:
Never mind. Any plans for that to change, or should I go about it the hard
way?  😁

panterarocks49:
I can add it

panterarocks49:
Will try to add it this weekend

mlava:
awesome, thanks!

mlava:
Hey @panterarocks49 
I think I've finessed the setup process and sorted the interval set and unset as requested.
Without access to an API function to check for a Shortcut, it will just pop up and alert reminding the user to create on if they are using shortcut mode. If they are using Menu item mode the div will be created and destroyed.
Cheers!

8bitgentleman:
Is there any way to make this a true circle or at least a horizontal oval? This squished vertical oval looks kind of strange
image

mlava:
@8bitgentleman will work on it, cheers
Is it working properly otherwise?

8bitgentleman:
Just installed it today and seems to work well! I'll play with it for a few days and let you know if there are any issues

panterarocks49:
@mlava Thanks for updating, I still haven't had a chance to add star page so I am going to merge this

panterarocks49 (approved):

mlava:
Thanks @panterarocks49 
Do you think you will add starred pages at some point?
I'll revise this once that's available.
Cheers!

PR #153, opened 2022-08-30
New Extension roam-toolbar

dive2Pro:


panterarocks49 (changes requested):
Why require instead of import? (just curious as I'm not a js expert)
https://github.com/dive2Pro/roam-toolbar/blob/67f5d4c12405d2e63d985937193e18e4ef308c9f/src/OperatorToolbar.tsx#L1

Can you delete extension.js from the repository? It makes it more obvious this extension is compiled (we run compilation on our servers so you don't need to commit it)

The gif in the readme needs to be the github url, Roam doesn't understand relative paths

Doesn't look like this file is used
https://github.com/dive2Pro/roam-toolbar/blob/67f5d4c12405d2e63d985937193e18e4ef308c9f/src/context.tsx

dive2Pro:
> Why require instead of import? (just curious as I'm not a js expert) https://github.com/dive2Pro/roam-toolbar/blob/67f5d4c12405d2e63d985937193e18e4ef308c9f/src/OperatorToolbar.tsx#L1

they are the same. already change it to import


> Can you delete extension.js from the repository? It makes it more obvious this extension is compiled (we run compilation on our servers so you don't need to commit it)

deleted.

>
> The gif in the readme needs to be the github url, Roam doesn't understand relative paths
> 

changed.

> Doesn't look like this file is used https://github.com/dive2Pro/roam-toolbar/blob/67f5d4c12405d2e63d985937193e18e4ef308c9f/src/context.tsx

deleted.

thanks for review. 

and i add more features to it

panterarocks49:
Thanks for updating! 

There is one more thing I forgot, your stripe account is still marked as restricted and you need to add a bank account or debit card to be eligible for payouts

panterarocks49 (approved):

panterarocks49:
Hey there seems to be an issue with your Stripe Account again, not sure what happened as I thought when I approved this is was working. 
image

dive2Pro:
seems like I forgot to activate the debit card.  It should work now.

Joshua Brown ***@***.***> 于2022年10月20日周四 03:10写道：

PR #154, opened 2022-08-30
reference path

paulovieira:
This is finally ready for review.

There is no stripe acount in the json. This is on purpose as the original work on this plugin was done by @azlen (Azlen Elza). I have contacted him by email about submitting the plugin and he said it was ok. I will contact him later to eventually split any possible payments.

More background here: https://github.com/Roam-Research/roam-depot/pull/48


panterarocks49 (approved):
Thanks for all of your work on this! Works great now

PR #155, opened 2022-08-30
Toggle Text uppercase/lowercase

8bitgentleman:
Requested on twitter. A simple right click menu plugin to toggle all block text to upper or lower case


panterarocks49 (approved):

8bitgentleman:
@panterarocks49  aw crap, looks like the name in the JSON was never changed and it has overwritten my "Table Import" extension in the store. My Table Import plugin, which had previously been doing quite well, is now showing as 0 and points to an incorrect extension. This is 100% my bad, apologies, hope the Table Import downloads can be restored 😨 
image

PR #161, opened 2022-09-01
Several bug fixes and improvements - Workspaces

mlava:
And allow [[page]] and ((block)) refs in definitions @panterarocks49 as per your suggestion https://github.com/Roam-Research/roam-depot/pull/128#issuecomment-1232272349


panterarocks49 (approved):
Nice, seems to work well

PR #176, opened 2022-09-03
Query Builder: Version 2022-09-03-16-14

dvargas92495:


panterarocks49:
I'm not sure if this is a new regression but if I accidentally query for something and have no results the component collapses
image

dvargas92495:
yes definitely a regression from the UI changes. Will address

PR #177, opened 2022-09-03
Workbench: Version 2022-09-03-17-44

dvargas92495:
Well It's finally here.

I decided to completely rewrite Roam42 and publish it as WorkBench to Roam Depot. All old files are in the `/legacy` file and are not used in the actual submission. They're there to give an idea of what the original code looked like, and to patch any emergencies that come up during the migration. This folder will be removed once WorkBench is live on RoamDepot for a few weeks.

There are probably several questionable instances of logic here - don't be afraid to be critical. I imagine that there are still things I need to clean up before it actually reaches the store.

### DIFF
--- a/extensions/dvargas92495/workbench.json
+++ b/extensions/dvargas92495/workbench.json
@@ -5,6 +5,6 @@
     "tags": [],
     "source_url": "https://github.com/dvargas92495/roamjs-workbench",
     "source_repo": "https://github.com/dvargas92495/roamjs-workbench.git",
-    "source_commit": "998336a992baff828a8bcf8d225d2a5105afcc52",
+    "source_commit": "dd86a56db0b2af2f3c81a85b2ae72ebf42487ef5",
     "stripe_account": "acct_1LGDNwQhxYPoO7v8"
 }
--- a/extensions/dvargas92495/workbench.json
+++ b/extensions/dvargas92495/workbench.json
@@ -5,6 +5,6 @@
     "tags": [],
     "source_url": "https://github.com/dvargas92495/roamjs-workbench",
     "source_repo": "https://github.com/dvargas92495/roamjs-workbench.git",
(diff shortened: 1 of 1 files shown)


panterarocks49:
Sounds good! I'll be posting reviews by feature, here is tutorials

Seems like you are missing some css
image

https://github.com/dvargas92495/roamjs-workbench/blob/998336a992baff828a8bcf8d225d2a5105afcc52/src/features/tutorials.tsx#L1250
This should be removeEventListener

panterarocks49:
Privacy Mode

Link is dead for the tutorial video on your website
image

Privacy mode command palette option still exists when it's disabled

Disabling privacy mode doesn't toggle it off, it just removes the css
I think if you just `if (active) toggle()` it will fix it

panterarocks49:
Format Converter

JSON options give me an error
image

Web view does not show up for me, a modal appears and then disappears, does that happen for you?

panterarocks49:
Deep nav

There are a lot of bugs here, not sure if it's worth mentioning since it's marked as under development

Could you remove the keyboard hints on disable? I think this is due to another bug but I ended up in a state where the keyboard shortcuts were stuck in the sidebar
image

panterarocks49:
Ok I think I am done with review

> There are probably several questionable instances of logic here

What did you have in mind as questionable?

panterarocks49:
> https://github.com/dvargas92495/roamjs-workbench/blob/main/src/features/formatConverter.tsx#L562-L564

Format converter web view looks like its doing remote script loading, just to render latex...

That is the one thing I had marked to review once the modal issue was fixed, since it ends up loading in an iframe I think it's ok

dvargas92495:
Had some users ask about WorkBench - any other points of feedback left for me to iterate on?

panterarocks49:
Nope, just had to go back and check over the changes

panterarocks49 (approved):

PR #182, opened 2022-09-05
Update Word Count - offer selection word count via kb shortcut

mlava:


panterarocks49 (changes requested):
`shift-alt-q` inserts a character on macOS, I think you need to do `e.preventDefault()` in your handler

mlava:
Thanks @panterarocks49 fixed in https://github.com/mlava/word-count/commit/e9761de1c56ff677acf64f0fcdeb157a750c0b1c

PR #186, opened 2022-09-07
Autotag: 2022-09-06

dvargas92495:


panterarocks49 (changes requested):
Is this the new home for page synonyms and unlink finder?
If so you might want to add some docs about how those work

unlink appears to be off but I am still getting some matches, is it off when the helper bar doesn't appear? or is it always on until I switch it off in settings? If it is on then it is not finding some of the matches
image

Also when I shut down unlink finder from the settings (see screenshot below, now there is no unlink finder button) it still shows the highlights, even after navigating to a new page
image

I am planning on force updating everyone early next week, I am going to wait to merge this until then to make sure everyone has the code for returning a function from onload

panterarocks49:
Another small thing that would be nice if it was fixed, unlink finder appears to not recognize that attributes are pages
image

dvargas92495:
Yes this will be the new home for those extensions - I've updated the docs accordingly

It should turn off after you toggle it off and I saw the bug - Added a fix and also transcribed unlink finder to typescript since the js version was driving me nuts lol

Will look into the attribute issue next

panterarocks49 (approved):

PR #196, opened 2022-09-10
New Extension, roam-hierarchy

dive2Pro:


panterarocks49:
There is a small problem with roam-hierarchy due to a recent change in Roam's API for create page. When I click on part of a hierarchy page that already exists I get an error. That is because we changed create page to throw an error when the page already exists. I am considering reverting this change because it's tripping a lot of people up, but if you want this approved before I do that you'll have to fix.

dive2Pro:
> There is a small problem with roam-hierarchy due to a recent change in Roam's API for create page. When I click on part of a hierarchy page that already exists I get an error. That is because we changed create page to throw an error when the page already exists. I am considering reverting this change because it's tripping a lot of people up, but if you want this approved before I do that you'll have to fix.

OK, thanks for pointing this out,  i just fixed it

panterarocks49 (approved):

PR #200, opened 2022-09-11
Fix Quick Capture with Todoist - broken label mode download

mlava:


danyalaytekin:
Hi @mlava, I noticed that tasks are no longer being deleted in Todoist after they're being imported and was wondering whether this was intentional.  Did you mean to comment out their deletion, assuming this is the line that was performing it?: 

https://github.com/mlava/quick-capture/commit/4f2a0048b47dfc26b0e4c6dc5c1f822aa1ed37d6#diff-19a9a3efb94bf1b84229bda71ab0ef5badcdbf34ba3a0c9e4152a82c569a21e6L361-R374

In the longer term, whether or not that line's change was intentional, perhaps deletion could become a configuration option:

```
Once import is complete:
    1. Delete on the Todoist side, or
    2. Tag on the Todoist side, or
    3. Do nothing
```

Thanks again for this extension - it's so useful.

mlava:
Hi @danyalaytekin 
No, it's an artefact of testing new code - I comment out the delete line to allow me to try things over and over, but I'm supposed to uncomment it before uploading. Sorry, and thanks for bringing it to my attention. 🙏

I kinda like this idea of yours. What purpose would doing nothing serve, though? I can see delete vs tag... but do nothing I'm not sure of the use case. It will be easy enough to add this feature.

Cheers!

mlava:
Hi @danyalaytekin 
New version awaiting review by RR team allows for the three after import options. Thanks for the suggestion!

mlava:
Hi @danyalaytekin this update has been reviewed and merged. Hopefully it's working as intended.

danyalaytekin:
Hey @mlava, thanks for your replies and sorry for going dark for a while there.  I really appreciate you implementing this (and your heroic output in general).  Will check it out.

PR #203, opened 2022-09-12
Update Numbered List - add refresh SB command and SmartBlock

mlava:


panterarocks49:
Why do you need to delete these blocks? https://github.com/mlava/numbered-list/compare/a6abb25ff43e59a90a39ced45a065783a69d04df..2f19ec26730968cd1ae32590088990fce0167cc7#diff-19a9a3efb94bf1b84229bda71ab0ef5badcdbf34ba3a0c9e4152a82c569a21e6R125

Is that the only way to get around this error?
image

mlava:
Yeah, it's an ugly hack. I couldn't figure another way...

panterarocks49 (approved):
Ok, I guess it is ok

mlava:
I'm not proud of it 🙄

PR #229, opened 2022-10-01
Roam Memo: Release v8: Fix Create API Regression

digitalmaster:
Just noticed that the page creation API now throws an error if you try to create a page that already exists: 
CleanShot 2022-09-30 at 21 27 11@2x

It also throws an error if you don't pass it a UID: 
CleanShot 2022-09-30 at 21 29 32@2x

This plugin (and i'm sure many others are broken as a result) 😬


baibhavbista:
Super sorry @digitalmaster 

We'd done better error reporting for the backend and it made sense to reuse the code so as to make it easier for extension devs to reuse their code too

I’ll merge your fix and write a message on the channel.

PR #234, opened 2022-10-04
New extension - zoteroRoam

alixlahuec:


panterarocks49:
I'm not seeing a config panel for this extension, but I do see in the code that you are loading from the new extension settings, did you forget to commit the config panel?
https://github.com/alixlahuec/zotero-roam/blob/main/src/setup.js#L220

image

alixlahuec:
> I'm not seeing a config panel for this extension, but I do see in the code that you are loading from the new extension settings, did you forget to commit the config panel? https://github.com/alixlahuec/zotero-roam/blob/main/src/setup.js#L220
> 
> image

So I'm managing the settings within my own React tree, rather than using a config panel. I'm using the Roam Depot structure to persist settings between sessions essentially, but then I manage the "editing"  aspect in a custom component.

I [had a chat with David about this](https://roamresearch.slack.com/archives/C02TMKXNVS6/p1659019556538949?thread_ts=1659016437.939479&cid=C02TMKXNVS6) when I was first looking around the extension API stuff, and it seemed to be the way to go

panterarocks49:
Sorry I missed that you had replied, I am still not understanding how to set the extension up.
How do I open your react tree to add my API key? 

I also tried following the [documentation](https://alix-lahuec.gitbook.io/zotero-roam/getting-started/api), which had me create a javascript block, but I was not able to see the icon in the top right that I should be able to see

image

alixlahuec:
> 

Right now there are two ways to install the extension - through roam/js, or the Roam Depot system.

### From roam/js

You need two things in your `roam/js` block:
- load the extension's code: https://alix-lahuec.gitbook.io/zotero-roam/getting-started/install (FYI v0.6 is the latest public release, submission above is for v0.7)
- create the `zoteroRoam_settings` object, with a `dataRequests` property that holds your API key etc: https://alix-lahuec.gitbook.io/zotero-roam/getting-started/api

Then refresh the graph so the roam/js code can run, and if the book doesn't show up it's either a syntax problem or a config problem (which should log an error in the console)

### From Roam Depot

No `roam/js` code is needed anymore, simply loading the extension as a plugin should make the book icon appear, then you can finish the setup by following the steps from the v0.7 documentation: https://alix-lahuec.gitbook.io/zotero-roam/v/v0.7/getting-started/api

(by default the docs show how things work for v0.6 because that's the latest public release, to see the v0.7 stuff you have to select that version in the dropdown)

panterarocks49:
Ah ok, I missed that their were updated docs! 

I found the issue, I had the wrong commit checked out, and extension.js and extension.css should be output to the top of the repository from build.sh, instead of inside of `dist/`

panterarocks49 (changes requested):
In addition to the [output directory](https://github.com/alixlahuec/zotero-roam/blob/13f3c62881b320467c0d286b096a653883d0dff7/dev/webpack.roam.config.js#L42) needing to be changed

The command palette commands added [here](https://github.com/alixlahuec/zotero-roam/blob/13f3c62881b320467c0d286b096a653883d0dff7/src/components/App/index.jsx#L92) need to be removed when the extension is uninstalled

I think that is all that needs to be changed and then we can merge this

alixlahuec:
> In addition to the [output directory](https://github.com/alixlahuec/zotero-roam/blob/13f3c62881b320467c0d286b096a653883d0dff7/dev/webpack.roam.config.js#L42) needing to be changed
> 
> The command palette commands added [here](https://github.com/alixlahuec/zotero-roam/blob/13f3c62881b320467c0d286b096a653883d0dff7/src/components/App/index.jsx#L92) need to be removed when the extension is uninstalled
> 
> I think that is all that needs to be changed and then we can merge this

🤦 Fixed both of those! ([PR here](https://github.com/alixlahuec/zotero-roam/pull/91/files))

panterarocks49 (approved):
great!

PR #240, opened 2022-10-07
Workbench: Version 2022-10-07-03-12

dvargas92495:


dvargas92495:
One of the new features in this diff is migrating the multi select extension, which with the release of Roam's latest feature, I could remove. Would hold off on reviewing until I do so

panterarocks49:
Is this good to review now?

dvargas92495:
Yes I removed it, should be set for review!

PR #247, opened 2022-10-10
Added the YouTube Extension

c3founder:
Create ccc-roam-yt


panterarocks49:
Marking here for anyone else looking at this review
The youtube extension dynamically loads the youtube Iframe API, but there is no other way to include it (npm wrappers still load it dynamically) so I think it's ok to do for this case

panterarocks49 (approved):

PR #255, opened 2022-10-15
Static Site: Version 2022-10-15-21-42

dvargas92495:


dvargas92495:
Removed the stripe account from this extension as well, since it operates with its own business model. In the future, it will become a product offered by SamePage

dvargas92495:
Going to close this - I think instead of ever launching this on Roam Depot, I'm going to start integrating publishing into SamePage

PR #264, opened 2022-10-21
Update - Footnotes - New features

fbgallet:


panterarocks49:
This is really cool!

I'm getting the same issue I did with the first review where I can't create a footnote, do I need to create the #footnotes block first or not?
image

The error says the parent block doesn't exist yet

My settings are the default ones that come with the extension

fbgallet:
I don't know why but the settings from the panel settings are not properly loaded when the extension is loaded in the graph for the first time in developer mode.
I've made the `onload` function an async one, and added some await before `extensionAPI.settings.set` and `extensionAPI.settings.get`, and it works now.

panterarocks49:
Weird, it is working now, thanks!

panterarocks49 (approved):

PR #302, opened 2022-11-04
Hypothesis: Version 2022-11-04-14-21

dvargas92495:


panterarocks49:
I can't get it to import any highlights, any ideas?

https://www.loom.com/share/de610a8107254fdf9af7df6bdf3dcb86

I have an API key and I don't see any errors in the console

Here is the url I was testing with https://www.reuters.com/world/us/republicans-close-us-house-majority-senate-still-up-grabs-2022-11-10/

PR #303, opened 2022-11-08
New extension: Ultraworking

mlava:


panterarocks49 (approved):
Small improvement that might be nice, allow for hitting enter to continue with the process

PR #306, opened 2022-11-09
New Extension: ESV Bible

tavon:
I've created a simple ESV Bible query/insert extension modeled after the Wikipedia extension.


tavon:
I forgot to update the source_repo. Fixing and resubmitting.

PR #307, opened 2022-11-09
New Extension: ESV Bible

tavon:
Submitting a small extension for querying and inserting passages from the ESV Bible (https://api.esv.org)


panterarocks49:
I'm not seeing your stripe account on our end, are you sure you set it up through Roam?

It should be the same account number as displayed on this page
image

tavon:
I've updated the stripe account to match the account setup through the Roam app.

panterarocks49 (approved):

PR #321, opened 2022-11-13
New Extension - Similar Pages

scottrblock:
similar pages uses machine learning and graph theory to help you find connections between ideas.

demo and overview - https://www.loom.com/share/d034cbf2909442e5bea4a6bca2a962f9

you can also see more on the [README](https://github.com/phonetonote/similar-pages), including a little about how it works (webworkers and indexeddb only, nothing being sent to a server). happy to answer any questions. I was originally working on this way back before roam depot launched, so excited to finally ship this!

### DIFF
--- a/extensions/phonetonote/similar-pages.json
+++ b/extensions/phonetonote/similar-pages.json
@@ -1,0 +1,10 @@
+{
+  "name": "Similar Pages",
+  "short_description": "uses machine learning and graph theory to help you find connections between ideas",
+  "author": "scott block",
+  "tags": ["machine learning", "natural language processing", "nlp", "graph theory", "universal sentence encoder"],
+  "source_url": "https://github.com/phonetonote/similar-pages",
+  "source_repo": "https://github.com/phonetonote/similar-pages",
+  "source_commit": "417480eb2bbe809d80ede5556dca035e633225a8",
+  "stripe_account": "acct_1LKwanQVF0QOKIg5"
+}
--- a/extensions/phonetonote/similar-pages.json
+++ b/extensions/phonetonote/similar-pages.json
@@ -5,6 +5,6 @@
   "tags": ["machine learning", "natural language processing", "nlp", "graph theory", "universal sentence encoder"],
   "source_url": "https://github.com/phonetonote/similar-pages",
   "source_repo": "https://github.com/phonetonote/similar-pages",
-  "source_commit": "417480eb2bbe809d80ede5556dca035e633225a8",
+  "source_commit": "61b8e870d130cf970d368ce603692f0a2e564572",
   "stripe_account": "acct_1LKwanQVF0QOKIg5"
 }
--- a/extensions/phonetonote/similar-pages.json
+++ b/extensions/phonetonote/similar-pages.json
@@ -5,6 +5,6 @@
   "tags": ["machine learning", "natural language processing", "nlp", "graph theory", "universal sentence encoder"],
   "source_url": "https://github.com/phonetonote/similar-pages",
   "source_repo": "https://github.com/phonetonote/similar-pages",
-  "source_commit": "61b8e870d130cf970d368ce603692f0a2e564572",
+  "source_commit": "3ff405740863f73621cc0c7bccaee5909d2e685e",
   "stripe_account": "acct_1LKwanQVF0QOKIg5"
 }


scottrblock:
I'll also note that I've only tried this on my m1 mac, it's possible I need to tune the webworkers a bit to not choke up on slower machines. This is all easy to tweak in `src/constants.tsx`, starting with `CHUNK_SIZE`

panterarocks49:
Cool!

Some UI feedback
The scatter-plot icon might look better as a minimal button so it matches the other ones in the navbar

the dialog's height is a bit too much for my screen and causes it to go offscreen and have a scrollbar
image


If I am understanding this correctly, you host embedding.js on your website, then download it and run it inside of a worker, which then downloads tensorflow from a cdn. This breaks one of our stated guidelines. Is there no way to include those inside of your script? Maybe by following one of the ways outlined [here](https://stackoverflow.com/questions/5408406/web-workers-without-a-separate-javascript-file)

scottrblock:
Hey yea, I can update the height and icon. 

That's right as far as how the webworker works. The code all runs clientside, and I saw @dvargas92495 does some webworker stuff, so I figured that was fine.

The cdn dependencies I realize are another story. I was hoping y'all would consider adding tensorflowjs to the scripts you include. It's great that you already have IDB package, so extension users can make use of clientside storage. I think including tensorflowjs would enable a lot of ml plugins. Being able to do everything clientside (idb + tensorflowjs) is really nice, especially with how expensive openai and similar are.

I'll look into the blob solution, but it looks like importing local dependencies to the inline script tag is still an open question on that stackoverflow link, so might not work

scottrblock:
I think this might do the job - https://gist.github.com/dtipson/d11f8068ae3a5f1907228fc2888b3705 😆 I'll give it a shot

dvargas92495:
The web worker stuff I had pre-roam depot, it was cut when I migrated the extension over

scottrblock:
ah ok, @panterarocks49 if webworkers in general are a problem let me know, but I think I need something to do this processing off the main thread. @dvargas92495 sorry to keep dragging you into this, but did you replace webworkers with lambdas? is there a place to see that in use? I'd prefer to use cloudflare if I need to use something serverless, but curious to see how you're doing it

dvargas92495:
The webworkers use case I was doing was scrapped entirely in favor of lambdas + backend API. The main reason I was using it was to do "background querying", which the backend API now enables.

Here's the repo of all the lambdas I currently have live - https://github.com/dvargas92495/roamjs-backend

scottrblock:
Thanks @dvargas92495. So far I'm having issues getting webworkers with dependencies working locally. I'll keep trying down that path for a bit, but another option would be rewriting the extension to use cloudflare workers/kv instead of webworkers and idb. I'd like to keep it all client side if possible

scottrblock:
I don't see a way to do an inline webworker while using an imported script, so even if it were created inline, I'd have to import tensorflow via the CDN. 

@panterarocks49 I'm not opposed to rewriting this with cloudflare, but I would say that currently its completely client-side, with the exception of those CDN calls (which cache well). If I rewrite it, I would be replacing client-side indexdb with serverless cloudflare KV (also caches well, but requires my cloudflare account to be running/billable). I would have to do this because I can't access indexdb from a cloudflare worker like I can a webworker. So on net, it'd be doing much more network requests, etc if I were to change it. I chose to build it all clientside because I thought the tradeoffs were worth it. Cloudflare serverless makes things easy enough to switch, but I think there's an argument for keeping it all local, and biting the bullet on CDN calls to load tensorflow. let me know what you think, would like to hear your thoughts before I start work on the cloudflare bit.

panterarocks49:
Web workers in general are not a problem, the problem is loading a script onto the users machine from your server. I'm open to adding tensorflow to our exposed libraries, but it would be a lazy loaded library, would definitely prefer to keep it all client side for users. Should I go ahead and add that?

panterarocks49:
hmm, on second thought, that won't work since web workers don't have access to the window object and that is where we expose libraries. If you know of a way we could expose that to a web worker let me know. 

I think the best solution for now is to go ahead and import tensorflow from a cdn in the webworker, but inline the code that you were hosting on your own website.

scottrblock:
Hey @panterarocks49 I just pushed some updates to get the webworker running locally and address the UI feedback. Appreciate the flexibility on allowing the CDN call in the webworker. I agree there'd be no way to have roam pre-load it and have the webworker access it

panterarocks49 (approved):

PR #325, opened 2022-11-15
New Extension - Sticky Headings

mlava:


panterarocks49 (changes requested):
You need to remove the command onunload https://github.com/mlava/sticky-headings/blob/a7dca197552d419f157e9107ebb5783b187544f0/extension.js#L6

I am getting an error when trying to turn it on (and then it doesn't work)
image

mlava:
> You need to remove the command onunload https://github.com/mlava/sticky-headings/blob/a7dca197552d419f157e9107ebb5783b187544f0/extension.js#L6
> 
> I am getting an error when trying to turn it on (and then it doesn't work) image

sorry Josh. Fixed the onunload.

Also, the issue was looking for headings that didn't exist. Have now rejigged so that it handles use cases where there is no heading of each type on the page.

Cheers!

panterarocks49 (approved):
No worries!

> Also, the issue was looking for headings that didn't exist. Have now rejigged so that it handles use cases where there is no heading of each type on the page.

Weird, I had headings on the page at the time, works for me now

mlava:
Yeah, that is weird. It's better that I implemented a fallback for pages without headings, so I'm glad for the nudge to behave better! :-D

PR #344, opened 2022-11-18
Update Word Count - bug fixes & handle multi-block select

mlava:


panterarocks49:
Hey I have to wait to deploy this because the new API was released yesterday and it takes awhile until everyone has the update

PR #345, opened 2022-11-18
New Extension - TODO Teleport

mlava:


panterarocks49:
Also have to wait on this one because it's a new API. This would be a good use case for adding to the native context menu so you can just right click to move todos instead of opening the command palette (or have both options)

mlava:
Yeah, that's a great idea. Are there docs for hooking into the context menu?

mlava:
Scratch that last... I found them. Is there any way to hook into the right-click menu that appears when click anywhere in the multi-select aside from the bullet?

Also, is there / will there be a way to programmatically unselect or turn off the multiselect mode once done? I guess I could just send a ctrl-m...

panterarocks49:
There is not an API for the context menu or for turning off, I can add them, do you want me to hold off on merging this until I add those? or merge and then add those features later?

mlava:
I think merge and add those features later. I wasn't sure if you would want
that menu being cluttered by plugins, but I think it would be convenient...

PR #347, opened 2022-11-19
Update - Find & Replace

fbgallet:
Major update, with a lot of new features...


panterarocks49:
By default, after loading the extension the toast doesn't appear because of
image
After changing the property in settings it fixes it, but the default should work

I think the highlight also doesn't appear until after I change the default selection

Search in page inserts "undefined" text into the search box
image

fbgallet:
It should be fixed now.

panterarocks49 (approved):

PR #348, opened 2022-11-19
New Extension: Startpage

8bitgentleman:
Lets the user set a custom page or block to load automatically on roam start instead of the Daily Notes page


panterarocks49 (approved):
Small thing you could fix in a subsequent PR
https://github.com/8bitgentleman/roam-depot-startpage/blob/2df7f5fcf19e669c8385cb04e2917e13ace529e5/extension.js#L26
offline graphs are /offline instead of app, you can check roamAlphaAPI.graph.type if it's offline or hosted

PR #349, opened 2022-11-20
new Theme "Native" dark

dive2Pro:


dive2Pro:
I had trouble creating a page using createPage and came up with the idea of using updateBlock instead.

panterarocks49:
What was the issue you were having with creagePage? we recently (a month or two ago) changed create page to throw an error if the page already exists, if that is the issue then you just have to check if the page exists first

dive2Pro:
> What was the issue you were having with creagePage? we recently (a month or two ago) changed create page to throw an error if the page already exists, if that is the issue then you just have to check if the page exists first

Yes, the problem I encountered has been posted here. https://roamresearch.slack.com/archives/C02TMKXNVS6/p1666967557591099

When I create a page like "A/B"  that doesn't exist, even if I call createPage it twice, no error is thrown and no page is created.

panterarocks49:
I'm sorry I missed that slack thread the first time, I am able to reproduce the behavior and I just released a fix. Can you update and check if it is fixed for you now? 
If it is fixed now, can you revert these changes, the way you are creating pages now will be slow on large graphs + might be buggy if there isn't an empty string block

dive2Pro:
> I'm sorry I missed that slack thread the first time, I am able to reproduce the behavior and I just released a fix. Can you update and check if it is fixed for you now? If it is fixed now, can you revert these changes, the way you are creating pages now will be slow on large graphs + might be buggy if there isn't an empty string block

thx, it's fixed now!

roam-hierarchy's changes has been reset

panterarocks49 (approved):

PR #355, opened 2022-11-22
Update: Startpage Fix for Desktop app and Hosted graph

8bitgentleman:
as mentions on the slack channel and https://github.com/Roam-Research/roam-depot/pull/348#pullrequestreview-1189059008


panterarocks49 (changes requested):
Looks good, but can you fix the conflict?

8bitgentleman:
Should be fixed @panterarocks49 , not sure what that was

PR #365, opened 2022-11-26
New extension: Table of Contents

mlava:


baibhavbista:
hey @mlava 

Initial impressions:

because of the extension, there is a large gap in my right sidebar before the content. I think it's because you are positioning the div there and later moving it. Please fix the issue
image

Another problem I had is that for the first heading, navigation to it did not work and I got the following console error
image

when i change right sidebar, size of toc box does not change and it can end up looking like this
image

mlava:
Hi @baibhavbista - thanks for reviewing and the initial feedback.
I've fixed the sidebar placement issue by placing the div at bottom of roam-main-body and then floating it up. It won't disturb sidebar content any more.
I haven't been able to reproduce the first heading link not working - is it simply an H1 with bold?
I think the third issue is also fixed, but will await your review.
Cheers!

baibhavbista:
Hey @mlava 
Thanks for fixing the issues 1 and 3
Looks like you removed the `**`-kind markdown remnants from the headings before showing as well, which is great!

Regarding 2, I think that the issue might be the method you use for scrollTo i.e. getting target via `document.querySelector('[id*=' + uid + ']')` will not work if the uid has an _ char. 
You might have to escape it somehow
Please check the error I get
image

mlava:
Thanks for the feedback @baibhavbista 
I've figured out the underscore in UID issue - the uid just needed to be surrounded by " ".
I think this is everything fixed now, and hope it will be ok to merge.
Cheers!

baibhavbista:
That's great @mlava 
After your fix, I confirm that it's working on my page too 
I'll merge the PR now

btw, one recommendation for future is to first get a diff between current state and new state and determine whether headings (either text or heading level) have changed before starting the work of rendering/adding the elems to the DOM

mlava:
Thanks @baibhavbista 
I presume you mean comparing before and after from the pullWatch? I actually don't know how to do that... any suggested links for me to learn from?

PR #373, opened 2022-11-29
New extension - Wordle

mlava:


baibhavbista:
Hey @mlava 

Some nitpicks:

1. One time, the request failed and I ended up with a block that said "Loading...". You might want to catch errors in fetch and then clean up

2. Personally I'd appreciate some hints/instruction on how to start playing. For example, the README.md fails to mention that you trigger it via the command palette. Similarly, after you trigger it, would love a way to hint me towards typing my word in the block. In that vein, after it is triggered, it would be great if you focused the user on to the block where they enter the word, using roamAlphaAPI.ui.setBlockFocusAndSelection  (doc here: https://roamresearch.com/#/app/developer-documentation/page/peR33ZsAb)

3. very small nitpick: sometimes the letters do not align. See INTER vs ASTER in the attached screenshot

image

mlava:
Hi @baibhavbista 
I've addressed all three issues and fixed each of them.
I hope this will be enough to be ready for publishing?
Cheers!

baibhavbista:
looks good @mlava , thanks

only concerns I have now are:
1.  the fact that you're scraping off of https://wordfinder.yourdictionary.com/wordle/answers/ and the extension will break if that website changes layout
2. the valid word list which is hardcoded and so will reside always in memory

but I think these are okay. I will merge

mlava:
Thanks @baibhavbista 
It's a risk, but I'll be sure to update if the source page changes format... I wondered about calling the word list only when required, but figured it would have to be in the extension the user installs anyway, so might as well just hard code it. I considered hosting the list somewhere, but then it would need to be accessed on demand which would be lots of fetching...

PR #408, opened 2022-12-20
HOLD NOT READY - New Extension - Art Museums

mlava:


baibhavbista:
Hey @mlava, trying to review all of your PRs today 😅

Some things:
1. Is the input supposed to be there at the bottom of the screen? 
image

2. for the requests (especially the advanced ones), you'd want to do the validation client side before calling the API. for example, here are some errors I got
image
image

3. art by dept name met seems to not work (it seems to create a "form" but either does not scroll to it or does not make it visible. As a result, nothing seems to happen, but when you then do a request that works and that requires a form, you see something like: 
image

4. similar issues as the other PRs where you'd probably want to remove the "Loading..." block if you run into errors, since you alert anyways
image

baibhavbista:
Oh yes, a suggestion more for general repo maintenance than anything, especially for the extensions that have a build.sh and use node modules, you want to have a .gitignore file and include the node_modules folder in that
This makes it such that you don't mistakenly add those folders to version control (git)

mlava:
Sorry @baibhavbista I guess it wasn't quite ready.
I'm missing the izitoast CSS so the form will be in middle of screen not at bottom :-D
Will fix issues and return.
Cheers!

mlava:
Closed to allow further work before resubmit

PR #409, opened 2022-12-20
New Extension - Word of the Day

mlava:


baibhavbista:
Hey @mlava , 
some nitpicks:

1. you want to clear the "Loading..." when you run into an error. In this case, I did get an alert of there not being an API key so it makes sense, but you'd still want to clear the block I think
image

2. It always links to https://www.dictionary.com/e/word-of-the-day/ and merriam webster equivalent. However, this means that I cannot go to the corresponding webpages for earlier day's pages
The links look pretty simple, maybe you could generate links specific to the day instead?
https://www.dictionary.com/e/word-of-the-day/posole-2022-12-21
https://www.merriam-webster.com/word-of-the-day/deleterious-2022-12-22
For these two, the end slug seems to be simply {WORD}-{DATE-STRING}, which should be trivial to generate and should work as long as they don't change their link format

3. In general, how reliable would you say rapidapi (or the APIs hosted on it) is? Have you used it extensively before?

mlava:
Hi @baibhavbista 
Thanks for the feedback.
I've fixed 1 & 2.
I've used the rapidapi platform a bit and not had any problems so far... Would be willing to just scrape the sources directly, but thought this would be easier :)
I hope this one is now ready to go...

baibhavbista:
looks good @mlava 
thanks for the changes, will merge now
no need to scrape the sources, this should be okay

PR #410, opened 2022-12-21
New Extension - Definitions

mlava:


baibhavbista:
@mlava , Some questions:
1. If the word already exists (as in the page `words/definitions/{word}` exists), do you rerun the request or directly open to the word's page in sidebar?

Idea for future update (Personal thoughts):
Most of the time, I don't want the definition to be saved in my roam graph (I find stuff like this clutters my graph). I'd preferably want to look at a definition in an overlay and if I want to save it, only then save it. 
My thinking of what the overlay could look like is macos's inbuilt thing which happens when you force touch on a word
image

Another thing that might be interesting in this vein is keeping a history of things I've searched and building a way I can revise and learn those words (I think amazon kindles have a way you can revise words you looked up the definition for and I think that could be helpful)

mlava:
Hi @baibhavbista 
Thanks for the feedback.
I've implemented a check so that if the word is already saved, it doesn't call the API it just opens the definition in the right sidebar. Good idea, thanks.
I've also created an option to have a toast with a save button to choose to save it, or a config option to auto-save all new definitions.
I think this is pretty much ready for an initial version, and plan to work on synonyms, examples, etc as well as prettying up the toast over time. Cheers!

baibhavbista:
Hey @mlava 

Sorry just a few things 😅

1. I couldn't click out of the popup/toast.. only way out was to click on "Save to graph" which kinda defeats the purpose of the toast 😅
It looks like making `close: true` should work?

2. can you update the gif in your readme to show the default dehavior?

3. While you're in the readme, maybe also mention the macos shortcut
OPT-SHIFT-d.

mlava:
Sorry, attention to detail is a bit lacking today.
I've fixed the toast close issue and updated the readme.
Thanks again @baibhavbista .

baibhavbista:
Great @mlava 
Will merge now

PR #411, opened 2022-12-21
New Plugin Datepick enhancer

dive2Pro:


baibhavbista:
Hey @dive2Pro 

You forgot to include either an extension.js or a build.sh in your repository
For more info, Please go through the readme.md for https://github.com/Roam-Research/roam-depot

In your case, I'd expect your build.sh file would need to include something like: 
npm install && npm run build

baibhavbista:
oh yes, you'd also want to remove the console logs

dive2Pro:
@baibhavbista thx, build.sh was added.

baibhavbista:
looks great @dive2Pro 
Will merge the extension

btw, a simple idea for future enhancement: allow people to select their own pages they want to use instead of "due". 
you might want to look into the extension API functions if you want to go down that route and have a panel for users to have settings for your extension : https://roamresearch.com/#/app/developer-documentation/page/y31lhjIqU

baibhavbista:
Hey @dive2Pro 
I tested your extension on a large graphs and it was really slow. took 30 seconds
image


I think the bottleneck is the command which counts all children. Could you try this instead? it should be much better

```
window.roamAlphaAPI.data.fast.q(
      `
      [
        :find ?uids (count ?child)
        :in $ % [?uids ...]
        :where
          [?page :block/uid ?uids]
          [?child :block/page ?page]
      ]
    `,
      ancestorrule,
      uids
    )
```

directly using :block/page should be orders of magnitude faster than using the ancestor rule which you were using

dive2Pro:
> Hey @dive2Pro I tested your extension on a large graphs and it was really slow. took 30 seconds image
> 
> I think the bottleneck is the command which counts all children. Could you try this instead? it should be much better
> 
> ```
> window.roamAlphaAPI.data.fast.q(
>       `
>       [
>         :find ?uids (count ?child)
>         :in $ % [?uids ...]
>         :where
>           [?page :block/uid ?uids]
>           [?child :block/page ?page]
>       ]
>     `,
>       ancestorrule,
>       uids
>     )
> ```
> 
> directly using :block/page should be orders of magnitude faster than using the ancestor rule which you were using (takes 350ms after the change)

wow, thx! This is much faster. 
and yes, User can set their own "due" page, which will be supported in next version

PR #416, opened 2022-12-23
new extension: website title parser

dragonforce2010:


panterarocks49:
You need a build.sh file that contains shell commands to build your extension instead of committing the extension.js file. Also it looks like your extension.css is leftover from the example repository

dragonforce2010:
> You need a build.sh file that contains shell commands to build your extension instead of committing the extension.js file. Also it looks like your extension.css is leftover from the example repository

gotcha, I will fix it, thanks

panterarocks49:
Is it intentional to send all of the paste data to your server? 
For example, if I paste "hello world  https://roamresearch.com test" then the entire string gets sent to your server, not just the url, this is also a bug, since it doesn't convert the url included

dragonforce2010:
> 

it's not intentional to send all the paste data to the server, it's a bug, let me fix it.

PR #419, opened 2022-12-24
Update todo teleport - fix available date range

mlava:


baibhavbista:
going to merge @mlava because you Dm'd about it
code looks safe enough

I'm curious why you're setting the MIN and MAX btw. were you running into perf issues?

PR #439, opened 2023-01-08
New extension: Block info

fbgallet:


panterarocks49:
the percent done is coming up as "undefined" for me

image

PR #441, opened 2023-01-09
New Extension: Autopin

sweenzor:
Thank you.


panterarocks49:
Hi, it looks like you forgot to add the `.json` extension to the `autopin` file. Could you update the file name? thanks

sweenzor:
Hey @panterarocks49, am I doing something obviously dumb to cause the tests to fail?

panterarocks49:
Our json parser is not very smart, it doesn't like the extra comma on your last key value pair I think

image

sweenzor:
> Our json parser is not very smart, it doesn't like the extra comma on your last key value pair I think
> 
> image
> 
> image

oh duh. thanks!

PR #449, opened 2023-01-14
New Extension: Self-Destructing Blocks

8bitgentleman:


panterarocks49:
Is it intentional to only run this when roam is restarted / opened?

8bitgentleman:
I had assumed that no one would just keep roam open but setting an update interval is probably a better practice. Should be updated now

panterarocks49:
You would be surprised! I pretty much only close it when I update

PR #454, opened 2023-01-15
New Plugin: Snapshot

dive2Pro:


panterarocks49:
Hey first of all think this is a cool extension, but there are a few issues.

Extension settings isn't really designed for holding the amount of data that will be put into it by this. Storing all of this data in settings will likely cause dramatic loading time increases and OOM errors, since settings for every extension is held in memory. It will also be somewhat expensive for us, since every page will be duplicated some number of times, and realtime database isn't cheap for storage. You will probably also eventually run into value size limits in realtime database.
Instead of using settings, I think you could store all of the data in a file in `roamAlphaAPI.util.uploadFile`, and in the settings map the UID to the file url. We don't yet have a deleteFile function but I could add one for you. let me know what you think

I also think "Page History" is a better name than "Snapshot"

I wasn't able to try it out because I am getting an error with building on my local machine, any ideas why?
I'm on node v16.18.0 (npm v9.2.0)
image

dive2Pro:
@panterarocks49 Yes, thanks for you feedbacks.

> I wasn't able to try it out because I am getting an error with building on my local machine, any ideas why?
I'm on node v16.18.0 (npm v9.2.0)

According to this https://nodejs.org/api/deprecations.html#DEP0128, i'm upgraded this vite-plugin, you can try it again, and if not work can you lower your node version, like the 16.13 version I'm using now (i'm use [nvm](https://github.com/nvm-sh/nvm) to do so)?

> I also think "Page History" is a better name than "Snapshot"


Because I want to support "Block History" at first too, but I might make a new plugin for this later.

Yes, I will rename this plugin to "Page History".

> roamAlphaAPI.util.uploadFile

I didn't know about this api before and didn't find it in the api documentation, can you tell me how to use it?

panterarocks49:
> I didn't know about this api before and didn't find it in the api documentation, can you tell me how to use it?

It is not yet available in the docs because there is a possibility we could change it and break the API, so I want to know who is using it. 

You can use it like this, it returns a promise with the download url
```
await roamAlphaAPI.util.uploadFile({file: new File(["foo"], "foo.txt", {type: "application/json"})})
```
You can just replace `foo` with your json string and `foo.text` with whatever you want to name it

dive2Pro:
@panterarocks49 Is it possible to add a property that controls the display of the uploading toast? 
like 
```
await roamAlphaAPI.util.uploadFile({
  file: new File(["foo"], "foo.txt", {type: "application/json"),
  toast: {
   show: false,
   message: '' // pass other toast's props here
  }
})
```

dive2Pro:
> We don't yet have a deleteFile function but I could add one for you. let me know what you think

I think this function is important, otherwise users may face this situation.


image

dive2Pro:
@panterarocks49 hi, any update on this?

dive2Pro:
Hi, or can I add the local version first?

baibhavbista:
@dive2Pro hey, just checking in. I think this is blocked until we expose being able to add new versions of uploaded files, yes?

dive2Pro:
@baibhavbista I don't know the reason. 

I didn't know when this new API would arrive, so I created a local version of the application using IndexDB. However, this plugin is still blocked until now.

panterarocks49:
@dive2Pro so sorry on the delay for this, I thought we were waiting on the deleteFile function. I just released `get` and `delete` file functions, and added a way to hide the toast for upload https://roamresearch.com/#/app/developer-documentation/page/S38S5jH2C

Would you like me to review the local version or wait to use the remote?

dive2Pro:
@panterarocks49 hey, I just uploaded the code with the new API - it is now open for review.

panterarocks49:
I am also getting a lot of errors in the console, but not sure why, is this ok or is something broken?
image

dive2Pro:
image

is this one ok?

---

> I am also getting a lot of errors in the console, but not sure why, is this ok or is something broken? image

This one is fixed; it was caused because I restored the first-level blocks twice.

panterarocks49 (approved):

PR #529, opened 2023-02-12
new extension: power chatgpt, bring chatgpt into the roam

dragonforce2010:
…l to make our work more efficient!

### DIFF
--- a/extensions/dragonforce2010/roam-power-chatgpt.json
+++ b/extensions/dragonforce2010/roam-power-chatgpt.json
@@ -1,0 +1,10 @@
+{
+  "name": "Power ChatGpt",
+  "short_description": "This extension is build upon the open ai ChatGPT model, you can now bring the ai into the roam now! let us chat with the today smartest model to make our work more efficient!",
+  "author": "Michael Zhang",
+  "tags": ["chatgpt", "ai"],
+  "source_url": "https://github.com/dragonforce2010/roam-power-chatgpt",
+  "source_repo": "https://github.com/dragonforce2010/roam-power-chatgpt.git",
+  "source_commit": "d53299d2593697ea6ebb2c38313a3fbf7a6f6652",
+  "stripe_account": "acct_1MI7MVQVLQeWv9Xo"
+}


dragonforce2010:
onhold this mr for now, I need to add a detailed readme

panterarocks49:
What is chat-pro-screen and why does it look like compiled code?

dragonforce2010:
> What is chat-pro-screen and why does it look like compiled code?

yes, it's compiled code. Let me optimize this pr.

baibhavbista:
Hey @dragonforce2010 

I saw that this extension sends the GPT requests to https://ec2-54-254-24-184.ap-southeast-1.compute.amazonaws.com and after that I assume forwards the request to OpenAI. 
 
I'm afraid we cannot approve this extension while it does requests to an unknown backend. If you change it such that it directly does a fetch on openAI's servers, that would make it okay.

Also, as @panterarocks49  already mentioned before, we cannot approve the extension while it has compiled code like `src/chat-pro-screen/sdk.js`

baibhavbista (changes requested):
Hey @dragonforce2010 

I saw that this extension sends the GPT requests to https://ec2-54-254-24-184.ap-southeast-1.compute.amazonaws.com/ and after that I assume forwards the request to OpenAI.

I'm afraid we cannot approve this extension while it does requests to an unknown backend. If you change it such that it directly does a fetch on openAI's servers, that would make it okay.

Also, as @panterarocks49  already mentioned before, we cannot approve the extension while it has compiled code like src/chat-pro-screen/sdk.js

dragonforce2010:
> Hey @dragonforce2010
> 
> I saw that this extension sends the GPT requests to https://ec2-54-254-24-184.ap-southeast-1.compute.amazonaws.com/ and after that I assume forwards the request to OpenAI.
> 
> I'm afraid we cannot approve this extension while it does requests to an unknown backend. If you change it such that it directly does a fetch on openAI's servers, that would make it okay.
> 
> Also, as @panterarocks49 already mentioned before, we cannot approve the extension while it has compiled code like src/chat-pro-screen/sdk.js

@baibhavbista all the comments have been addressed, please review it, thanks

baibhavbista (changes requested):
sent some comments via slack @dragonforce2010

baibhavbista:
Adding my notes for the comments I had here too, for future reference:

- something weird happened when I ran `./build.sh` locally. It produce 100+ js files in the directory
    - image
    
- they'd added an API token of their own. Seems to not be working. Since the token is in a public github repo, would not be surprised if a scraping bot found it. Might be a good idea to remove the default API token and force users to use their own key
    - image

- when sending the "messages" in the chat/completions request, wrong messages are being sent (the first 8 are being sent while the last 8 should be)
    - image
        -  here what it’s doing is only keeping the first MaxContentMessageSize messages on each request
while you probably want to be sending the last MaxContentMessageSize messages

- I think max_tokens should be customizable. Turns out it was has been hardcoded in the source code to 200. I had been wondering why my requests were getting cut off in the middle
    - image
   
- want to use extensionAPI’s addCommand for the "Roam GPT" command instead of the roamAlphaAPI one
    - image

- nitpick: spelling mistake in “try again”. might be problematic since it is a message sent to openai as well and may give wrong answers
    - image

- I’d also expect most users of Roam are english users (+ best experience when interacting with chatGPT is in english I think) so maybe placeholder text should be in english as well?
    - image

- slight nitpick: what about only showing the parts inside triple quotes using codemirror?
    - image

- another codemirror issue: when a line is highlighted, it’s white by default which is a problem given that font is also mostly white. Compare this with the earlier screenshot
    - image

If I have more comments, will be sending them via slack

baibhavbista:
@dragonforce2010 reminder that this PR is still in requested changes state

baibhavbista:
Marking this as a "draft PR" in order to not confuse & accidentally release

Please change back to a regular PR after looking at the requested changes @dragonforce2010

PR #530, opened 2023-02-12
new extension: power css pack

dragonforce2010:


dragonforce2010:
onhold this mr for now, I need to add a detailed readme

extensions/dragonforce2010/roam-power-css-pack.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "Power Css Pack",
+  "short_description": "This extension added a lot of fancy css utilities to further enrich your roam",
+  "author": "Michael Zhang",
+  "tags": ["css", "power pack", "extension", "roam"],
+  "source_url": "https://github.com/dragonforce2010/roam-power-css-pack.git",
```
panterarocks49, line 6: Can you update this to not have the .git at the end?

dragonforce2010: sure, will do it:)

dragonforce2010:
can anyone review this pr?

baibhavbista (changes requested):
Hey @dragonforce2010 

Again, sorry for the delay in reviewing the extensions

I have a couple of comments:

1. I _really_ don't like that you're creating new css classes starting with `rm-`. The reason is that we (the roam team) are using that convention for our CSS classes. If you also create new classes with the same naming scheme, it can easily lead to issues in the future if we add classes named the same as yours (mistakenly). Would like it if you renamed all of the new css classes starting with `rm-`, like `rm-bg-level`, `rm-grid`, `rm-border`, etc

2. Would prefer that the extension only loads the css on the `onload` and cleans up the CSS pm the `onunload` function

3. Personal review: I was trying to use your extension since some of the css classes looked useful. However, right out of the box, it installed a theme which I personally did not like. Would prefer if it did not include any theme.
image

dragonforce2010:
@baibhavbista thanks for the comments and please check out the latest pr, all comments have been addressed.

baibhavbista (commented):
Hey @dragonforce2010 

thanks for taking care of the first 2 things I mentioned in the earlier review: https://github.com/Roam-Research/roam-depot/pull/530#pullrequestreview-1401727373

I saw that you did not do number 3 (disable rainbow theme). My rationale for that was that I'd expect most users would not want that theme. Were you not able to factor it out of the extension?

dragonforce2010:
> 

I have removed rainbow.css from my code in my current pr, can you check again?

baibhavbista:
@dragonforce2010 oh sorry I mistakenly called it rainbow.css

By the "theme"/"rainbow theme", I meant this thing where it auto sets the color of each level of blocks in the graph. If you want it, I think you could make it a setting, or even better, only activated when a css class like .css-rainbow-children is used. My feeling is that most users will not want it on by default. 

image

The POV I'm looking at this extension through is that as a user, I can use the css classes provided by this extension as I need it and ONLY then. Since it's a power pack, I wouldn't expect to change default behavior of any code. If I wanted default behavior, I'd expect that from a theme extension. Hope that makes sense

dragonforce2010:
> @dragonforce2010 oh sorry I mistakenly called it rainbow.css
> 
> By the "theme"/"rainbow theme", I meant this thing where it auto sets the color of each level of blocks in the graph. If you want it, I think you could make it a setting, or even better, only activated when a css class like .css-rainbow-children is used. My feeling is that most users will not want it on by default.
> 
> image
> 
> The POV I'm looking at this extension through is that as a user, I can use the css classes provided by this extension as I need it and ONLY then. Since it's a power pack, I wouldn't expect to change default behavior of any code. If I wanted default behavior, I'd expect that from a theme extension. Hope that makes sense

good catch, comments has been addressed, as per your suggestion, the font level color will be only activate when putting `#.css-level-color`

baibhavbista:
thanks for fixing, @dragonforce2010. everything looks good to me now, will publish

PR #545, opened 2023-02-20
update: rename extension

dragonforce2010:
rename extension name from "Website Title Parser" to "Roam Power Link Parser"


panterarocks49:
Can you change the name of the json file back? it breaks our system, you can change any of the keys in the file, but changing the name of the file changes the internal ID and existing users won't get updates (I think this is why the build on the PR is failing too)

dragonforce2010:
> Can you change the name of the json file back? it breaks our system, you can change any of the keys in the file, but changing the name of the file changes the internal ID and existing users won't get updates (I think this is why the build on the PR is failing too)

got it, I will do it, thanks.

PR #555, opened 2023-02-21
Update Todoist TM - allow auto sync of new comments, subtasks, rename…

mlava:
… tasks etc


mlava:
And move to extensionAPI...

PR #563, opened 2023-02-24
Small fixes: Find and replace

fbgallet:


panterarocks49:
Why not change all of the [commands](https://github.com/fbgallet/roam-extension-find-replace/compare/915a245dac64a13fdc13e3bb269698636c891cd8..5d43507b33740220983c497f3d415e258743d422#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556R3867)?

panterarocks49 (approved):

fbgallet:
@panterarocks49 because I thought that all the commands didn't need a shortcut, but in fact they will appear in "Ungrouped Extension Hotkeys", which is no better.

PR #573, opened 2023-02-28
New extension: Auto-DNP templates

mlava:


8bitgentleman:
Isn't this already a part of smartblocks?
image

mlava:
Thanks, @8bitgentleman I've been using my own roam/js script for so long that I'd forgotten that.
However, it's only a single workflow allowed, whereas my extension allows a different workflow for every day of the week or weekend/weekday workflows.
I think this allows more flexibility for the average user, although if you're a Smartblocks power user you can clearly make things show on specific days of the week.
Cheers!

8bitgentleman:
Fair enough!

PR #587, opened 2023-03-01
New Plugin: Switch+

dive2Pro:


baibhavbista:
Wow I really like this extension @dive2Pro !!!

Some comments from the usage:
1. In the vanilla Roam theme (light mode), in some places, it feels like the colors aren't what they should be. Maybe you didn't notice because you seem to use a dark theme predominantly
image
2. When I was not in a page (I was in the DNP), i tried the shortcut and I thought maybe the extension was not working. I found out that it only works inside of pages when I opened the console and saw the error. Maybe a toast/other dialog alerting the user that "Switch+" only works when you are on pages?
3. When the right sidebar is closed and you get to right sidebar mode by doing cmd+shift+p and then pressing "r:", it shows "clear sidebar" only (which I don't think works). I'd expect it to either open the right sidebar itself (like what cmd+shift+u does) or to have the option "Open sidebar"

Here's a loom (also contains some personal opinion which may or may not be right)
https://www.loom.com/share/9894b3ec300149af802a8101db68d681

dive2Pro:
@baibhavbista Could you please review this plugin again? I have pushed the new code to fix the issues you suggested

baibhavbista (commented):
Hey @dive2Pro

So I used your extension in a large graph and it was pretty slow (visible 3-5 seconds after I press cmd+shift+P), even on pages that are not that large

I had some suggestions for perf https://www.loom.com/share/7d3bfe37e30d4f6f823c0d08c928da66

Do watch it and let me know if you want to release the extension before or after perf improvements

(the reason I'm a bit concerned about perf is that I assume your extension will be a hit for power users, and power users generally have large graphs)

dive2Pro:
Hi @baibhavbista, thank you very much for your work once again. Your suggestion is invaluable as always.

Here are the results after my deep dive into the performance optimization process:


> all tested on this huge page https://roamresearch.com/#/app/roam-book-club-2/page/JaA5hSAXX

| Before | After |
|---|---|
| 2000+ms | 100+ms |


1. The main driver of this performance improvement process is the `{:block/refs [:block/uid :block/string]}` query
2. getAllChangesWithin2Day is only taking 70+ms, which is acceptable. 
3. splitting `[(> ?time ?start_of_day)]` into a `Array.filter` has significantly decreased performance.

Regarding the performance, I am also unsure about the extent of the graph that you have tested. Therefore, I welcome you to test the performance again. Cheers!

baibhavbista:
Thank you so much @dive2Pro, it's much more usable on the large graph now!!  ❤️

I did run into an error on a graph when trying to load the switch view
image

I think it might be originating from here (maybe it is called when the ref hasn't yet been initialized?
image

I think the extension is okay to merge after this small fix

baibhavbista (changes requested):
https://github.com/Roam-Research/roam-depot/pull/587#issuecomment-1535914888

dive2Pro:
@baibhavbista  thx again!
all **Some blocking suggestions** should be fixed.

**Some non-blocking suggestions/bugs**

1. I delayed the render phase for 250ms, this may help
2. Now, only entering the sidebar mode  will triggers the sidebar info fetch. 
   It takes time as I use Roam query to get the title or string, so if there are many windows it could take a some time.
3. open in sidebar was directly use roam api to do so, can you made a video for me to check this?
4. fixed
5. fixed
6. 

> why not use [roamAlphaAPI.ui.rightSidebar.removeWindow](https://roamresearch.com/#/app/developer-documentation/page/fl-GzIEJM) here?

Because I thought some windows may not have `block-uid`?

image

image

baibhavbista:
Hey @dive2Pro 
the extension is superb now ❤️
and performant on even large graphs and pages

I'll report some bugs here but these can wait until the next version probably. I want to get this out to people as soon as possible 😅

1. In some searches (especially in ref mode), I have blank lines. Screenshot below
image


will be adding to this as I find out more

PR #593, opened 2023-03-05
new extension: implement 23 roam themes

dragonforce2010:


dragonforce2010:
can anyone review this pr?

baibhavbista:
Hey @dragonforce2010 
Really sorry for the late review, had been swamped with other issues 😢

Regarding this PR, I'm not really sure if having **23** separate extensions is the correct approach here. I think having them separate would make sense if this was a theme store, but since this is an extension store first and foremost, this feels like an antipattern.

Maybe you could try an approach similar to Alexander Rink's Roam Studio?

Some advantages of having a single extension rather than a lot of extensions:
1. Easier for user: they can change theme by just going to extension settings, they don't have to install new extensions.
2. Better for visibility of the extension since having a single extension consolidates the downloads and higher number of downloads means it appears higher up in the list of extensions and so leads to a higher chance of new people installing the extension
3. Easier to fix in case you have bugs in the extensions (see below)

One bug I found with at least some of the extensions is that they do not remove the theme css when you uninstall the extension. What you want to do is remove the CSS on the `onunload` function. 
(This issue illustrates the pros of a single extension approach. If you had a single extension, you could fix this in one place. However, if you have 26 extensions, you have to fix the issue in all of them and then all 26 of them have to be reviewed again too)

dragonforce2010:
@baibhavbista good suggestion, I have merged all of the theme extensions into one extension, please help review again.

baibhavbista:
Thanks @dragonforce2010 , I will take a look today

Btw sorry to be a bit pedantic but can you separate out the two extensions into separate PRs? (we like having changes to one extension in one PR)
Also, a personal feedback on the name: "Roam Link Parser" sounds to me like the extension parses Roam links only

dragonforce2010:
yeah sure, I have just removed `extensions/dragonforce2010/roam-website-title-parser.json`

baibhavbista:
@dragonforce2010 did a quick prelim review of the extension [dragonforce2010/roam-power-themes](https://github.com/dragonforce2010/roam-power-themes)

Some notes:
1. couldn't figure out how to switch extensions. You'd probably want to setup a settings panel and have a select setting to select the theme
Documentation: https://roamresearch.com/#/app/developer-documentation/page/9wcwH8rH0
Code example of a select setting: https://github.com/panterarocks49/settings-panel-example/blob/main/extension.js#L37

2. Using extensionAPI's settings will also help you store the settings, so you don't have to store it in localstorage. We generally discourage people using localStorage and indexedDB in extensions while there are native solutions we offer / can offer. One of the many reasons (like user security) is that using the native stuff has advantages like syncing across user's different machines. But yeah, in this case, I think you might not have known about settings panel via extensionAPI. If you use that and remove localStorage usage, it should be fine

3. Small suggestion on applying the CSS and removing it. Instead of setting value of `document.body.className` (which can be brittle, say Roam adds a class there, or say other extensions do), could you do it in the way it is done in Roam Studio (screenshot below)
image

baibhavbista:
Oh no @dragonforce2010 
you do not want to delete the file `extensions/dragonforce2010/roam-website-title-parser.json`, that would remove the extension from the marketplace
What you want to do is undo the diffs.
Maybe you should just create a new PR for both of the extensions?

dragonforce2010:
gotcha, let me create a new pr for this

baibhavbista:
closing this PR since we now have #669

PR #630, opened 2023-03-29
New Extension - Automatic DNP

mlava:


baibhavbista (changes requested):
Hey Mark @mlava , sorry for the late review. 
I made a loom for my review
https://www.loom.com/share/7a56f06e4762481aa7585238bb750e1a

mlava:
> Hey Mark @mlava , sorry for the late review. I made a loom for my review https://www.loom.com/share/7a56f06e4762481aa7585238bb750e1a
> 
> edit: oh yes another thing, on every day change, you also somehow need to get new currentDate and use that (maybe a setInterval which checks if the date has changed every hour or so? You need this because many people keep their roam graphs always open (even at night) and if you only run it on onload you will not run the code for subsequent days

Thanks @baibhavbista.
The extension checks the page on hashchange https://github.com/mlava/auto-DNP/blob/main/extension.js#L106 so that should cover anyone who leaves RR open all night.
Given the hashchange function, would you still prefer I change to a pullwatch? I've been using this for a couple of years as roam/js script and it fires every time for me. I guess I don't usually open the DNP in right sidebar...

baibhavbista:
Hey @mlava 

> The extension checks the page onload and on hashchange https://github.com/mlava/auto-DNP/blob/main/extension.js#L106 so that should cover anyone who leaves RR open all night.
Will it? I'm not sure. If the user was in the DNP when the clock strikes midnight, a new page for the new day will be created, but will the hash be changed?
Edge case stuff like this (and the opening today's daily notes in sidebar) is why I was leaning more towards the pullwatch approach.
(however, even if you use pullWatch, you'd still need to have some other async thing that removes yesterday's pullWatch and adds today's pullWatch)

> if I did make it a pullwatch function, I presume you only mean for the current date? And I clear it once it's fired for this day.
Yes that is exactly what I mean. On the extension load, you might need to check the today's page exists (via a normal q or pull) and if it is not, then add the pullwatch

> The way this script works now, it won't trigger the template on looking at a future DNP, only if it matches the current date. This means users can put todos, notes etc in their future DNP, and then when it gets to that day the DNP template will be appended. I don't really want to put the template in if I create tomorrow's page which fires the pullwatch for tomorrow, I only want the template to appear once I get to tomorrow. Does this make sense?
No this makes a lot of sense, you should definitely keep this behavior. But you'd still keep this behavior with a pullwatch too (just watch for the uid of `currentDate`, a var which you already have)

> Also, thanks for the nudge to overcome my laziness and deal with the nested trees. Much more neat: https://github.com/mlava/auto-DNP/blob/main/extension.js#L217
Haha thanks, yes it does look a lot neater. When I saw that code, it surprised me a lot because I didn't immediately see what it did. Even after realizing what it did, I tried to think of some way to do it in a loop before realizing that one had to employ recursion (I'm blaming it on me not doing much Javascript the past few years XD)

baibhavbista:
@mlava did you want to work more on this or should I release?

mlava:
Yes, will fix before release. Have been under the pump outside of roam and didn't have time this week. Can I let you know when I'm further along?

baibhavbista:
Sure @mlava, no hurry, do "Request Review" and mention me in a comment here when it's ready

baibhavbista (commented):
already was in a state of requested review. Do request again after you think it's ready, @mlava

mlava:
@baibhavbista I think I've got this right now.
It creates pullWatch and then when fulfilled will check and create a pullWatch for tomorrow if needed. Also has an hourly interval check to see if the DNP has been created and will removePullWatch if now exists. 
I think this now meets your queries. Please let me know how you find it! Cheers,
Mark

mlava:
PS @baibhavbista 
I amended the regex to check for relemma pages as well :-D
Also, have created a check for DNP in right sidebar so that it populates the template there as well.
Cheers!

baibhavbista (commented):
sorry @mlava I might've been unclear earlier
can you checkout this loom? https://www.loom.com/share/50146f8f44b948e3be5e1c388dca09f1

mlava:
@baibhavbista 
Ok, I think I see what you're getting at now. You're proposing a dramatic simplification of how the template is triggered. I've been trying to cover hashchanges and all the ways it might be triggered. You're suggesting a very simple mechanism: onload check if today's DNP exists, if it does then apply the template, if not create pullwatch. Once template is applied, set interval to next day to apply the next one. That will need a page exists check as well, or perhaps aim to apply a few minute after midnight so RR has had a chance to create the new day's DNP. That should work for as long as RR remains active in browser. If RR is closed, the process starts again on next RR extension onload. I can then remove all the hash watches etc, and just use onload and setinterval as the triggers.
Is that about right? I originally wrote this as a roam/js script and converted it to an extension. I think I've been stuck on having to have the DNP on a screen somewhere for this to work. But you're suggesting it can be performed in the background, whether the DNP is open in the main window/right sidebar or not. The template can be inserted quietly and be ready for when it's required.
Is this above all correct? Do I understand this now? I hope so cause I think this makes sense. 🤞

mlava:
@baibhavbista 
Ok, I _think_ I've got it working now as you've suggested. New update available for review.

In this version:

- onload it checks if today's DNP exists
- if it does, it prints the template
  - then, it sets an interval to midnight and will check at that time for the next DNP
  - if the DNP exists it prints and sets a new interval
  - if it doesn't, it creates a new pullwatch
- if it doesn't it creates a pullwatch

Once a pullwatch fires, the DNP template is printed and the process starts over for tomorrow.

I think it's now more simple and should capture all uses. Please let me know what you think.
Cheers!
Mark

baibhavbista:
@mlava thanks for bearing with me and going along with the fixes! :D 
I think it's okay for release now, will be merging it

Btw, when using it, I realized some things that might be useful: (or might not be, these are just some thoughts I had)  
1. a command to trigger the template for the day (in case the user deletes the blocks and wants them back) or in case the extension does not work for some reason (someone who's looked into the code or that knows how roam extensions work can do the same by reloading Roam, but most users will not know that I assume. Therefore, I think having a command makes sense)
2. A command to (force) trigger the template for the daily notes page I'm currently on 
  i. sometimes I might want to apply template on an older day (tbf I could do this by applying the roam/template using ;; too)

PR #643, opened 2023-04-06
Export formatter new version 20230405

ryanguill:
I am having trouble with the readme in my roam depot page, any hint about what I might be doing wrong would be helpful!


panterarocks49:
I'm not sure, the error says the file doesn't exist, so maybe it failed to upload on the previous update, let's try merging this one and seeing if there are still issues

panterarocks49:
Oh I see, we are expecting a file named `README.md` whereas yours is `readme.md`, could you change?

PR #648, opened 2023-04-13
Small fix: Page References counter

fbgallet:


fbgallet:
Conflict should be resolved

PR #664, opened 2023-04-25
fix: Search+

dive2Pro:


baibhavbista:
can you elaborate in this PR on what the fixes are, @dive2Pro ?

dive2Pro:
@baibhavbista The entire difference in the store file was due to differences in line endings: CRLF versus LF.

this pr fix 2 things:
1. When we selected some blocks and opened search+, press 'escape' accidentally deleted these blocks. To prevent this, I simulated a click event to blur the selection when opening search+
2. https://roamresearch.slack.com/archives/C04H4739402/p1682280436382219. I'm trying to fix this reported bug, as I suspect it's caused by searching for a non-existent block.

PR #667, opened 2023-04-27
Important fix: Page References Counter

fbgallet:
@panterarocks49 some users have experienced app freezing with the previous version, on page with a large amout of linked references. This commit should fix the issue.


baibhavbista (approved):
Hey @fbgallet 
Looks good for the most part. Will merge

Btw, I found some weird behavior when using it in the "show only on hover" mode. 
For example: 
1. Sometimes the counts would not show up
2. It won't work for pages with other page names in the title

I believe those can be fixed later in a separate update, particularly because this update has already been sitting around for a long time (sorry for that btw)

PR #669, opened 2023-05-01
new extension: roam power themes - refactored pr of #593

dragonforce2010:
This is refactored pr of #593

### DIFF
--- a/extensions/dragonforce2010/roam-power-themes.json
+++ b/extensions/dragonforce2010/roam-power-themes.json
@@ -5,6 +5,6 @@
   "tags": ["markdown"],
   "source_url": "https://github.com/dragonforce2010/roam-power-themes",
   "source_repo": "https://github.com/dragonforce2010/roam-power-themes.git",
-  "source_commit": "45a8e7d9709a5947f5f6862abc8959a4e1561454",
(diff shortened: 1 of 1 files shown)


baibhavbista (changes requested):
Hey @dragonforce2010 

1. Using this extension, at least on my machine, some themes don't look that usable
image

2. Instead of window.roamAlphaAPI.ui.commandPalette functions, can you use extensionAPI.ui.commandPalette functions? Documented here: https://roamresearch.com/#/app/developer-documentation/page/JIAxMS1kJ
Because you're using the old one, the hotkeys are grouped under "Ungrouped extension hotkeys". See screenshot below:
image

3. currently, every time the user reloads Roam, the theme gets set to darkage
I think that is because of this line in the code
image
What you want to do instead is fetch the theme user has set using `extensionAPI.settings.get` and then if that does not exist, only then set the default theme


4. if you do it as I mention in 3, this interval won't be required either. You want to have a function that both sets the theme setting and also loads the theme by adding name to classList.
image


I then think you might be able to remove all of these
image

baibhavbista:
Hey @dragonforce2010 
I'm still not sure about having commands for each theme, but going to publish this nonetheless (since it has been in review state for so long)
image

everything else looks good

PR #671, opened 2023-05-02
Website Title Parser: refactor code and update readme

dragonforce2010:
refactored the code: using lambda function to provide a more stable service and update readme

### DIFF
--- a/extensions/dragonforce2010/roam-website-title-parser.json
+++ b/extensions/dragonforce2010/roam-website-title-parser.json
@@ -5,6 +5,6 @@
   "tags": ["markdown"],
   "source_url": "https://github.com/dragonforce2010/roam-website-title-parser",
   "source_repo": "https://github.com/dragonforce2010/roam-website-title-parser.git",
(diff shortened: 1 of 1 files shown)


baibhavbista (changes requested):
Hey @dragonforce2010 

Can you do one thing?
Do not apply the extension for Roam links and firebase links (since those do not have title anyways)
So, anything that contains/begins with
https://firebasestorage.googleapis.com/v0/b/firescript-577a2.appspot.com/
or
https://roamresearch.com/

Would also make this safer in general because if the server was compromised, user's files would not be compromised

Edit: Another thing to make it work properly when user switches blocks, please checkout loom below: https://www.loom.com/share/1f617d2557f844cf970de23ecb4eb23a

baibhavbista:
hey @dragonforce2010 looks good to me, please do remove the change to mlava/aot and then I'll merge / release

btw since old backend is down, I'll release this right after, but later, do take a look at doing the other things I mention in the loom above, I think it would make the extension more usable

PR #688, opened 2023-06-06
Roam Studio: Additional small fixes

rcvd:


panterarocks49:
Hey @rcvd overall we can merge this if you want but I have a couple of comments about the code that could help clean it up / reduce the number of lines. If you don't have time to clean up the code right now just lmk and I'll merge tomorrow

Is there a reason you store settings in your own var instead of just using `extensionAPI.settings.get` [here](https://github.com/rcvd/magic-tags/blob/1b35d04962188177066e63b1a9f76649cc985787/src/index.ts#L116)

Instead of [copy / pasting](https://github.com/rcvd/magic-tags/blob/1b35d04962188177066e63b1a9f76649cc985787/src/index.ts#L208) a bunch of the same type of setting, you could use a `for` loop to generate it, that would reduce the number of LOC by a lot

rcvd:
@panterarocks49  It would be great if you can merge it right now, so I can get feedback from the users. I will look into your suggestions next weekend. Thanks Alex.

rcvd:
@panterarocks49 I've changed the creation of the settings to a for loop. Not sure what you meant by "using extensionAPI.settings.get" instead of my own variables.

panterarocks49:
Thanks! I meant instead of `word = settings["mt-" + id + "-word"].toString();` you could do `word = extensionAPI.settings.get("mt-" + id + "-word");` and then you wouldn't have to use your own object to track the settings. The select settings component automatically sets that var when the user changes the component.

rcvd:
Awesome, good to know.

PR #714, opened 2023-07-06
Create Capture for Roam Research companion extension

mlava:


baibhavbista:
Some background info (from conversation with @mlava  on Slack)

> I have my flutter capture app available on Google Play Store (https://play.google.com/store/apps/details?id=com.lavercombe.quickCaptureRoamResearch) and in review on ios App Store. The companion roam depot extension is at https://github.com/Roam-Research/roam-depot/pull/714
I don't suppose the companion extension could be reviewed soon? It's only 30 odd lines of code and two CSS rules. They are intended to work together. In future, I will allow config for the app via the extension.

baibhavbista:
Hey @mlava, the code for the extension does seem to create a page as it should
image


However, I don't understand the authentication key uuid and how you're planning on using that. Could you elaborate?

Also, you might want to link to the Google Play Store and the App Store from the first block in the page too? Would make it easier for people to install if they haven't yet

mlava:
Hey @baibhavbista 
That's a great idea. I've updated to link to Google Play Store, and will do so once ios App Store is live (still in review).
The uuid is retrieved by the app and stored as a shared_preference. https://github.com/mlava/quick_capture_roam_research/blob/main/lib/auth.dart#L204
Once auth is completed with app the page is updated with a green highlighted text saying the app has been authenticated with the companion extension.
The utility of the companion extension page is really going to be for settings re tags to apply, where to add the notes (DNP vs named page) etc.
Are you on Android or ios. If Android you could install and try it out. I still need to work out a first launch workflow to allow user to add credentials from RR graph etc.
Cheers, Mark

baibhavbista:
looks okay, will merge / release

PR #719, opened 2023-07-11
Update tweet-extractor.json

8bitgentleman:
@baibhavbista thanks for the suggestions, I've implemented them here and it seems to have fixed the settings bug.


baibhavbista:
looks good to me, will merge @8bitgentleman, thanks !

PR #729, opened 2023-07-16
New Extension: graph-radar component

8bitgentleman:
roam/render component to show the last x edited pages in a graph. Useful for multiplayer graphs and graphs not often visited

image

### DIFF
--- a/extensions/8bitgentleman/graph-radar-component.json
+++ b/extensions/8bitgentleman/graph-radar-component.json
@@ -1,0 +1,10 @@
+{
+    "name": "Graph Radar Component",
+    "short_description": "Roam Research component to show a list of the last pages edited.",
+    "author": "Matt Vogel",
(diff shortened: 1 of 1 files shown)


panterarocks49:
You should check out our `{{x-activity}}` component, it's pretty similar to this one

Your code works and this is a pretty small component so I'll approve it but for next time I think you should read about reagent [form-2 / 3 components](https://github.com/reagent-project/reagent/blob/master/doc/CreatingReagentComponents.md). Generally creating an atom or any of our `roam.datascript.reactive` should be done in either the form 2 or 3 initialize state.

Also `(radar-page i)`
https://github.com/8bitgentleman/roam-depot-graph-radar/blob/bbe5aa261661c2add82521117d5521be69a1d213/src/component.cljs#L57
Should be `[radar-page i]`, reagent evaluates and caches the form for you.

PR #730, opened 2023-07-16
New Extension:Tag-cycle Component.json

8bitgentleman:
A roam/render component to cycle through a block ref'd list of tags
example2
example3

### DIFF
--- a/extensions/8bitgentleman/tag-cycle-component.json
+++ b/extensions/8bitgentleman/tag-cycle-component.json
@@ -1,0 +1,10 @@
+{
+    "name": "Tag Cycle Component",
+    "short_description": "Roam Research component to cycle through a list of tags or pages with a click.",
+    "author": "Matt Vogel",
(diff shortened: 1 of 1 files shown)


panterarocks49:
Couple of small things

Clicking on the button to switch tags selects the text for me (only happens if I click through it pretty fast)
image
I think you can just add `user-select: none;` to your css of that component

I was confused for quite a while because in your demo you hit the button to cycle from `#[[]]` to the first tag in the list, but for me that would not work and I had to manually type in the first tag in my list, is this a new bug since you made the GIF?

panterarocks49:
@8bitgentleman checking in, did you see my review? If you do not intend to fix I can merge as is, I was mainly tripped up by the difference in having to type in the first tag in the list in order for it to switch to another tag vs what I saw in your video

8bitgentleman:
Sorry! this totally fell off my radar. I'll try and take a look this week

PR #754, opened 2023-08-14
Workbench: Version 1.2.6

dvargas92495:


panterarocks49 (approved):
It would be nice if the docs links opened up in a new tab like instead of the current one @mdroidian @dvargas92495

mdroidian:
@panterarocks49 ✅🙌 https://github.com/Roam-Research/roam-depot/pull/775

PR #763, opened 2023-08-28
New Extention: Metamind-Roam

farhaanbukhsh:


farhaanbukhsh:
@baibhavbista Can we merge this plugin? It's been some time now.

baibhavbista (changes requested):
Hey @farhaanbukhsh  I went through the code for your extension and my impression is that the idea is that the Roam depot extension has 2 purposes:
1. generate "log" and "index page" (I assume these are special Metamind specific pages that you use for metadata regarding what to publish) 
2. a button to publish the graph to MetaMind.

First thing is that we generally check if the extension is working properly before merging it. Can you get me an account so that I can test? (I did register via "Submit your email" but I didn't get an email and I assume the process after that is manual?)

Second, Can you add your homepage (https://www.metamind.network/ ? ) to the README.md? There are a lot of things named MetaMind so it will spare prospective users (exploring from Roam Depot) from having to google and go through a bunch of results. This is actually what I meant earlier when I told you to update your README

Third and most important, the way you're storing roam-graph-tokens is not recommended. I'll explain why below, but can you not ask the user to enter the token in your app (and save it securely there) and only have buttons which trigger the two actions in the Roam extension.
Okay, now I'll go into the issues with storing the roam-graph-tokens that way (i.e. storing it in localStorage)
1. Other extensions or rogue roam/js scripts that users might have enabled could read that key in localstorage and that would be a breach of the user's data
2. localStorage is only stored by the browser. So, storing it in localStorage means that the user cannot use your extension from other clients (this could be other browsers, desktop app, or other computers)
What most services do is ask the user to submit the roam-graph-token in their UI and then securely store it there.

farhaanbukhsh:
@baibhavbista  I have fixed the comments, please let me know if I can clarify anything here.

baibhavbista (changes requested):
@farhaanbukhsh sorry for the ping on a saturday

Here is my review https://www.loom.com/share/eef211e8ac1748ae9de796ffbbdb03c0
I have a bunch of requested changes

baibhavbista (changes requested):
@farhaanbukhsh 
Please watch the following loom: https://www.loom.com/share/fb4a61cb947a4ea2abb0f1c5c2031119

I think I found one serious issue with your logic, and have other suggestions too (no hurry to fix the extensions)
Let me know what you decide to do about the "Blocker to merge" and then I can merge this PR and release the extension

image

PR #787, opened 2023-10-12
Publishing a "Roam-Preview_Window

mehboobelahi05:


mehboobelahi05:
Roam Preview Window Plugin

panterarocks49:
@mehboobelahi05 just one small issue

https://github.com/mehboobelahi05/Roam_Preview_Window/blob/a577eb6d03a8d90c2ab99b7398b2a8fcdca03b4a/extension.js#L89
`setInterval` needs to be called in your `load` function, then record the interval id and call `clearInterval` in `unload`.

PR #795, opened 2023-11-08
Create roam _preview_window.json

mehboobelahi05:
add new commit id.
Made the requested change of clearing interval when unload in the extension.js


mehboobelahi05:
Added the interval function inside the load in the extension.js file. 
Updated the new commit id in the .json file.

panterarocks49 (approved):

PR #800, opened 2023-11-23
New extension: Speech-to-Roam

fbgallet:


fbgallet:
This version is ready to be published

baibhavbista (changes requested):
@fbgallet A loom video reviewing the PR: https://www.loom.com/share/78ad51726ecd4876b9b491ad5b4a6af3

Also, the loom video with safari issues:
https://www.loom.com/share/6ac85f90845c431a93c8791f3d4bff33

baibhavbista (changes requested):
@fbgallet looks like this on my device, pretty sure this is a bug
Other issues seem to be solved  (no double clicks required for example)
Other than that, does not work on the ios mobile app, yeah?
https://github.com/Roam-Research/roam-depot/assets/23657255/980d09f1-7c85-42c9-878c-c5a18029c5a8

fbgallet:
@baibhavbista I think this time it's the right one!
Yes, currently it doesn't work on mobile app, only on mobile browser.

baibhavbista:
@fbgallet 

I think 1 is the main blocker for release
If noone else has had 2, would be okay with releasing and then trying to figure out in what cases that happens

image

PR #807, opened 2023-12-09
V2 Search+, inline search+

dive2Pro:


baibhavbista (changes requested):
Some comments @dive2Pro 

1. First thing that jumped out to me is that you should add a debounce to when the search text changes (so that it only searches after say 200ms after the text has stopped changing): That feels like a good low hanging fruit - should make it feel snappier to users

2. I think there might be quite a bit more optimization possible.
However, on first read through, the source code for the extension is pretty daunting.
Could you make me a video showing the main parts of the code - that take up the most compute and memory usage, and maybe I can suggest ways to make it more efficient? Would appreciate some insights into what stuff you store/cache in memory, how, when, how many datascript queries you run, etc

3. Btw regarding the inline searches, are you sharing any info/computation between different instantiations visible at the same time?
I’m on a large graph and I think multiple instantiations slow things down by a lot (lets say I’m navigating to a page with 3 inline searches, I have to wait a minute for the app to be responsive)
Also very very easy to run into Out of Memory issues. Maybe try not front-loading all the info at start for inline searches?
image

dive2Pro:
@baibhavbista thanks for review

1.  The refresh is triggered by the input blur event, not the input change event. I think there are no issues with the frequency of this trigger?
2.  Yes. 
    1.  Initially, when there is {{search+}} on the page or when search+ is opened for the first time, a full data query (roam api query all block & page) is performed, and the data is processed and saved in memory.
    2.  Each time search+ is reopened or a requery is triggered on {{search+}}, the data modified within the past 48 hours is queried and processed.
    3.  In {{search+}}, every time the filter is modified, the filter data is saved to the block's props.
    4.  During each {{search+}} rendering, the filter from the block's props is read.
3.  How large is the graph you are testing? After my modifications, there should be a significant performance improvement. You can try again.
    1.  after:



1.  before:

dive2Pro:
> Btw regarding the inline searches, are you sharing any info/computation between different instantiations visible at the same time

YES, {{inline search+}} and search+ both share the same cached data.

dive2Pro:
I think the performance has been improved, Here are the results:

and you can see at the end of the video, the memory cost will decrease.

https://github.com/Roam-Research/roam-depot/assets/23192045/580cbfaa-eb3c-4e9d-b50c-090fe7b05c8a

baibhavbista (changes requested):
Thanks for replying to my concerns and making memory footprint smaller, @dive2Pro !

Just one last thing, your PR had an error in the build process. Could you fix it? (should just require adding dep to package.json). 

After that, I think this extension update PR should be okay to merge

```
ERROR in /home/runner/work/roam-depot/roam-depot/checkout/hyc+roam-search+/src/inline-search/result/SearchResultGridView.tsx
./src/inline-search/result/SearchResultGridView.tsx 15:21-35
[tsl] ERROR in /home/runner/work/roam-depot/roam-depot/checkout/hyc+roam-search+/src/inline-search/result/SearchResultGridView.tsx(15,22)
      TS2307: Cannot find module 'marked-react' or its corresponding type declarations.
```

image

baibhavbista:
@dive2Pro just checking, the extension is ready now, yeah?

baibhavbista (approved):
lgtm will release now

baibhavbista:
@dive2Pro just released this update for Search+
we'd talked about making new videos for this. When you do that, do open a new PR and let me know and I will expedite the review
Btw, another thing, could you add some code which cleans up these `recentlyViewed` and `searchHistory` arrays if they get too long?
image

PR #815, opened 2023-12-21
Roam Power Preview - fixed some issue

dragonforce2010:


baibhavbista (approved):
This is an update I requested myself - users had a bug where extension would stop working after switching between DNP and page views

PR #817, opened 2024-01-03
Nautilus – the first pull request

tombarys:


tombarys (commented):
New commit id: I have added Quick Start video to Readme.md

baibhavbista (changes requested):
@tombarys I made a loom about some requested changes
https://www.loom.com/share/d347125d1ef94de3821b75db486573ee

baibhavbista:
@tombarys 
Block refs when changed are still not propagating. 
What I meant was something like this, this should work (you can remove `extract-ref`, new replacement fn is `str-with-resolved-block-refs`)

```
(defn str-with-resolved-block-refs [{:as block-ent :keys [block/string block/refs]}]
  (reduce (fn [string ref-ent]
            (str/replace string (str "((" (:block/uid ref-ent) "))") (:block/string ref-ent)))
          string
          refs))

(defn get-children-strings [block-uid]
  (r/with-let [*get-children-atom (rdr/pull
                                   [{:block/children [:block/string :block/order {:block/refs [:block/string :block/uid]}]}]
                                   [:block/uid block-uid])
               *children (r/track eval-state *get-children-atom)]
    (map str-with-resolved-block-refs
         (->> @*children
              (sort-by :block/order)))))

```

tombarys:
Hi, Baibhav, the reference extraction works now, thank you very much! I hope I was successful removing the setting.json file.

PR #822, opened 2024-01-08
New extension: Disable Auto Pair

alfred8jones:
ready to be published.

### DIFF
--- a/extensions/alfred8jones/disable-auto-pair.json
+++ b/extensions/alfred8jones/disable-auto-pair.json
@@ -5,5 +5,5 @@
     "tags": ["double", "pair", "brackets", "auto"],
     "source_url": "https://github.com/alfred8jones/RoamAutoPair",
     "source_repo": "https://github.com/alfred8jones/RoamAutoPair.git",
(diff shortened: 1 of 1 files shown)


baibhavbista (changes requested):
@alfred8jones extension looks simple enough and should be okay to merge

Except for storing state in localStorage, we generally recommend against that because
1. localStorage is not namespaced on a extension-by-extension basis (so other extensions might override yours)
2. localStorage is not synced across user's devices/browsers/etc. 

Why are you not using `extensionAPI.settings.get`? I see that you're doing a `extensionAPI.settings.set` but not using it anywhere. 

I think the extension is okay to be merged after making it use extension settings rather than local storage

baibhavbista:
@alfred8jones 
In addition to the localstorage issue above, one other comment on this extension - 
There seem to be a lot of commits (everytime you update the README.md maybe? there are 165 commits as of right now, which makes me think this is somehow automated). 
Can you squash all the commits on this `roam-depot` PR so that only a single commit will be added to the git history of the repo on merge?

alfred8jones:
will come back fixed in a few days

PR #823, opened 2024-01-08
Roam Power themes - Added Theme Config Center

dragonforce2010:
Added theme config center to better optimized the user experience
1. user can access the power theme extension from sidebar and toolbar
2. user can preview and quick switch between themes and have a good preview of the theme UI effect


baibhavbista (changes requested):
Some small comments:

1. The Drawer/Config center looks nice. 

2. However, I do not like that it opens up by default on page load/reload

3. The config center's icon should probably watch with the theme, yeah? 
image

4. Recommend adding the ability to disable / not show the icon in the nav bar (maybe a boolean in the extension settings?)

PR #835, opened 2024-01-28
Update Speech-to-Roam to v.2

fbgallet:


fbgallet:
@baibhavbista i've pushed some fixes on small bugs introduced in the last updates (on focused blocks especially, and the way block selection is handled),

and following your request, i've added commands for command palette and other small changes, see changelog here: https://github.com/fbgallet/roam-extension-speech-to-roam/blob/main/CHANGELOG.md

PR #844, opened 2024-02-02
Oura Ring: Version 1.0.0

dvargas92495:


baibhavbista (changes requested):
@dvargas92495 (maybe cc @mdroidian too?)
Looks like one of the major changes is that you're hitting 
"https://api.samepage.network"
instead of 
the Oura Ring API "https://api.ouraring.com/v1"

I feel like that is a big change (given that you're passing the user's Personal access token with the request). What is the reason for that?

If at all possible, let's not change the behavior since users might have already installed the extension thinking it doesn't send data anywhere else.
If you cannot do without doing that, one thing I'd expect is a change to the README.md to reflect that. Maybe something like "As of vX, uses samepage network for reason Y"

mdroidian:
@baibhavbista 
Oura's V2 API was rejecting the frontend with CORS errors, so we decided on using samepage's backend.

I've added the warning to the README 👍

dvargas92495:
The endpoint is also open source: https://github.com/samepage-network/samepage.network/blob/main/api/apps/oura/get.ts

@mdroidian let's add a link to the endpoint to the README, with also an explainer that we are simply rerouting the request to Oura to comply with their security policies

baibhavbista:
Okay CORS issues strike again XD

@dvargas92495 yeah I think being as transparent as possible would be best here. Do ping me when the thing you mentioned is done

mdroidian:
@baibhavbista updated!

PR #851, opened 2024-02-15
New Extension - Crosswords

mlava:


baibhavbista (changes requested):
Some requested changes @mlava 

1. The main problem with the extension implementation right now is that it uses LocalStorage 
    - separate key for every crossword instantiation, so can, over time, take up a significant portion of the [[LocalStorage]] limit (5 MB per domain?)
        - image

    - the reason we are saving guesses in localstorage seems to be that the underlying library mlava is using is doing that
        - maybe we can set `useStorage` to `false` in `CrosswordProvider`?
            - See https://github.com/JaredReisinger/react-crossword/blob/main/src/CrosswordProvider.tsx#L84
                - image

            - this would not only solve localStorage issue, but might even make sync code not required (the code which sync between data blocks in Roam & localstorage)
2. I would rather the "Crossword Definition" and "Crossword Guesses" blocks (the data blocks?) be children of the actual block rendering the crossword (with the rendering block having `open` = `false` so that user hopefully does not see the data blocks)

mlava:
Thanks @baibhavbista 
The underlying library doesn't have an option to use anything other than localstorage.
I think that setting useStorage to false means that no data will be save and as soon as the user navigates away from that page the progress is lost.
I will see if there is anyway to hijack to the save to localstorage, but I think it will be quite hard to do as it's baked into the original library.

baibhavbista:
Adding resources from our Slack conversation here too:
1. I would first check if there was an API method to get the Guesses data
2. Else/additionally, Could you not use onCellChange  and/or other callbacks of CrosswordProvider ?
image

3. Another potential building block: you’d probably want to use/base some code off of serializeGuesses  and deserializeGuesses  in the utils (it seems to already be exported, so you should be able to use them no problem)
image

4. Do try to use the above suggestions. But if it turns into too big of a project, another thing I think would be okay with is if your extension code cleaned up the localStorage entries (for example, when the react crossword component is being unmounted, and maybe also when the extension loads)

baibhavbista:
sorry just getting to this now @mlava 

Thanks, it looks good now. especially the removal of keys which begin with `roam_research_nytcrossword_` on load and unload makes me feel much better about this

Will release now!

mlava:
thanks @baibhavbista all good :)

PR #869, opened 2024-04-07
Nautilus: Changing the way how it installs itself 

tombarys:
First PR is just for testing purposes.
- checking for `running` in extension
- checking for component code update
- removed handling the `nautilus-disabled`

### DIFF
--- a/extensions/tombarys/nautilus.json
+++ b/extensions/tombarys/nautilus.json
@@ -5,6 +5,6 @@
   "tags": ["time-blocking", "agenda", "tasks", "planning", "productivity"],
   "source_url": "https://github.com/tombarys/roam-depot-nautilus",
   "source_repo": "https://github.com/tombarys/roam-depot-nautilus.git",
-  "source_commit": "08152fd8807663f19b95cd98e88b07b74f4fd38e",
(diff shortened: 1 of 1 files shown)


tombarys:
Dear Baibhav, as for the code: it seems like this install/uninstall implementation is much better than the previous one. No more unnecessary updates of render strings across graph, no unnecessary updates of the code block (until different from the one loaded from plugin). It works well including the backwards compatibility in case users leave nautiluses previously in the disabled state. All tested in many scenarios. Thank you very much for your idea!

No readme changes yet. Will do soon. Just check if there is another possible issue which I cant see. Thanks!

tombarys:
Ok, I still have a problem with the ((roam-render-nautilus-cljs)) block erase. When the block itself is erased, it sometimes leaves the whole content of the block inserted in place of the reference, which is really strange.

baibhavbista:
it looks good now @tombarys 
Thank you a lot for your patience 😅

I sent over a future request in DM and am including it here too:
> one thing for the next PR/update, in the settings screen, could you mention somewhere that changes made will start working only for newer/future invocations of the template (maybe you can find a better way of phrasing this?)
  Reason being that I myself got confused as well: I made a change and it was not working. Then I realized that I need to reapply the template

PR #883, opened 2024-06-01
Update code-block-copy.json

8bitgentleman:
Fix for an error when copying codeblocks


baibhavbista:
@8bitgentleman 
Probably not an issue with the new update, but an issue I found for large code blocks with your extension

https://www.loom.com/share/960ef49dab06419ebc0f510f86509a3c

P.S. Video was meant to have no sound. Somehow it captured speaker too. Sorry for the sound, please mute 😅

8bitgentleman:
@baibhavbista that makes sense, just updated the extension. I opted to keep the old solution for inline code blocks because people often have many inline code blocks in one roam blocks

PR #884, opened 2024-06-03
New extension: Full Calendar

fbgallet:


matt-vogel:
Playing with this today, it opens up a lot of possibilities. Could you save the user's view preferences when they change them? ie I find myself always including the DNP and switching to the week view instead of the month. Instead of doing that every time, can whatever setting I use be saved and loaded back up each time?

8bitgentleman:
Playing with this today, it opens up a lot of possibilities. Could you save the user's view preferences when they change them? ie I find myself always including the DNP and switching to the week view instead of the month. Instead of doing that every time, can whatever setting I use be saved and loaded back up each time?

8bitgentleman:
A tall order but maybe the ability to render the calendar inline could be nice, something like `{{calendar: week view:other settings}}` so it could be included in daily templates set to a certain view

fbgallet:
@8bitgentleman thanks for the suggestions!
These were ideas for v.2 but i've immediately implemented auto-save of user's view preferences (separately for the main page and the sidebar), I will consider a {{calendar}} block component later, since it need some tests, not sure to implement it.

panterarocks49:
Hey @fbgallet this extension looks awesome! I haven't gotten into the code yet but here are some usability issues I'm having

There is a bug with the filter search bar that happens sometimes but I can't quite nail down when. In this case I think I freshly installed the extension on a new graph and then opened calendar and switched on daily notes. Then none of the popovers worked until hiding and showing the calendar again. This same bug was causing all but TODO/DONE to not be shown in the calendar as well. 
https://www.loom.com/share/6c6240d866e94211a38837c9bfa3fe7e

It's confusing to me that all linked references to a day show up in the calendar, shouldn't the selected filters also apply to the linked references? Related to this, the "not tagged" option will hide them but then it is described as "children of #calendar" , as far as I can tell the "not tagged" filter has no effect on children of the #calendar block

I think you are using localstorage to set some of the view / tag settings but (particularly with which tags are selected) I think this could be confusing because it will carry over across graphs, but not across devices. You can use our [extensionAPI settings to set](https://roamresearch.com/#/app/developer-documentation/page/uS4xIy062) any key value pair you would like and it will sync across devices and be per graph.

If click on an item in the calendar and then remove the tag, it doesn't remove it from the calendar 

It would be convenient to just add a tag here from the search
image

panterarocks49:
btw, not saying you need to add all features of those to merge, the first bug I described was pretty bad though, I thought the extension was broken

fbgallet:
Hi @panterarocks49, Thank you for your feedback!

- I had never encountered the 1st bug before, but I understood where it could come from and it should be fixed now.
- I made sure that the "not tagged" filter is more consistent and works the same way for linked refs and for the blocks under the calendar tag.
- I completely replaced the use of localStorage with the extensionAPI (I hadn't thought about it but I wasn't sure if we could save data that doesn't correspond to extension settings, yet it was quite obvious, thanks for the suggestion!).
- Regarding the fact that an event remains displayed even if we have deleted a tag because of which it was displayed, it's a choice: it allows further interaction with the event if a false manipulation has been made.
- I plan to add the possibility to directly add custom tags from the search bar in a future version.

fbgallet:
I've fixed an important issue with listener on calendar button (lost when opening/closing left sidebar !)

panterarocks49 (approved):
Alright code looks good to me and I can't reproduce the 1st bug anymore, merging now!

PR #886, opened 2024-06-07
New Extension - Chess Puzzle of the Day

mlava:


baibhavbista:
@mlava is it just me or is the link to the API not working?
https://rapidapi.com/KeeghanM/api/chess-puzzles
image

baibhavbista:
Oh wait maybe I have to create an account & token via Lichess instead?

https://lichess.org/account/oauth/token

Even when doing that, it only seems to partially work. Some metadata has been fetched but the chess component is erroring out I think
image

baibhavbista (changes requested):
@mlava requesting changes because the extension is not working for me. Please check comments above

mlava:
> @mlava is it just me or is the link to the API not working? https://rapidapi.com/KeeghanM/api/chess-puzzles image

It appears to just be you:

image


Shouldn't need any account at Lichess.org, just the RapidAPI one.

mlava:
I've just tested on several browsers that have never browsed to lichess.org and with the rapidAPI key it works beautifully...

image

This is the same puzzle you're trying to load, as far as I can see...

Did the RapidAPI signup work ok? @baibhavbista

baibhavbista:
Oh interesting
Wonder why rapidapi is not working for me then
That link still does not work for me, even after I've signed up
image
Any idea what the cause might be @mlava ?

mlava:
> Oh interesting Wonder why rapidapi is not working for me then That link still does not work for me, even after I've signed up image Any idea what the cause might be @mlava ?

it's like the page doesn't even load...
maybe try another browser?

baibhavbista:
good news, it's now working for me :D
image

mlava:
very strange!

baibhavbista:
I know right. No idea why it didn't work before!

The extension looks good to me @mlava , will release

PR #887, opened 2024-06-10
New extension: Contribution Graph (Github-like)

Dagulf795:


baibhavbista (changes requested):
@Dagulf795 looks like a really fun extension. I left some comments requesting changes below:

- ### Suggestions/requested changes
    1. Would prefer a setting to make it available via a command (in cmd palette) instead of something that always takes up space in my navigation bar
        - resources for it
            - addCommand
                - `extensionAPI.uid.commandPalette.addCommand`: https://roamresearch.com/#/app/developer-documentation/page/XThfc9LJ6
            - settings panel for the extension
                - `extensionAPI.settings.panel.create` and `extensionAPI.settings.get` and `extensionAPI.settings.set`
                    - https://roamresearch.com/#/app/developer-documentation/page/9wcwH8rH0
    2. add gif or screenshots in README.md
        - you already have a video in the Roam slack. You could use that, or personally, I'd like it if it was not that zoomed out (i.e. font size larger)
        - sample: https://github.com/8bitgentleman/roam-depot-clear-sidebar/blob/main/README.md
    3. package.json cleanup 
        - can you remove the commands you're not using from `package.json`
            - I think these exist because this repo started from a clone of `phonetonote/similar-pages`
            - 
        - can you remove dependencies you're not using?
            - `mathjs`
            - maybe others too?
    4. when multiple years, show the year too
        - see example of a graph which has gone on multiple years
            - 
    5. Why do you have server side rendering turned on?
        - it seems most of the code complexity comes from the fact that you're getting svg from the module (I assume this is because of the ssr setting?). Would it not have been easier if using client side rendering?
        -

Dagulf795:
Hi @baibhavbista - changes now made. See [here](https://github.com/Dagulf795/contribution-graph/commit/1b6d1cc2af8792f0d62bff7f41db7144e6ed938e)

Thanks!

extensions/Dagulf795/contribution-graph.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "Contribution Graph",
+  "short_description": "Display your Roam activity",
+  "author": "Dagulf795",
+  "tags": ["activity", "contribution graph", "contribution", "GitHub"],
+  "source_url": "https://github.com/Dagulf795/contribution-graph",
+  "source_repo": "https://github.com/Dagulf795/contribution-graph.git",
+  "source_commit": "09c8eaf5e39d8c3dee789b6607ce9fff40f6f0e2"
```
baibhavbista, line 8: @Dagulf795 it looks like you solved most of the issues I asked, thank you

Just one issue, **after making an update to the source code for the extension, you need to update the `source_commit` here in the JSON**

To make sure the extension matches how it works locally, I also recommend checking your extension using the PR-shorthand (Mentioning this because testing via this would have caught the issue above)

image

You'd use that via the "load extension from URL" option in Settings > Roam Depot > developer Extensions (screenshot below)
image

baibhavbista (changes requested):
@Dagulf795 you need to update the source_commit to the version of the repo you want to build. Please check my comment below with clairification

baibhavbista:
Hey @Dagulf795 
Super sorry for the late reply and re-review!! I was a bit busy with other work and this ended up slipping through.
I found a few issues:

1. the default value for the "Show Contribution Graph Button" is off, but by default it is shown
    - you seem to have init code for this but it does not run because the check for when this runs is incorrect
        - required fix looks like
            -  image

2. sometimes I was able to get it to this state (can't really find a reproduction for this bug, it seemed to happen more regularly when the contribution button was the first thing I pressed)
    - image

I do know that this PR has been almost ready for such a long time, so if you'd prefer for me to release it as it is, and you fix them later, I am totally fine with that too. Please let me know

And again, sorry for getting back so late!

baibhavbista:
@Dagulf795 checking in about this again. Do you want me to release with the issues (& later fix it) or fix the issues before release?

PR #889, opened 2024-06-12
Update Workspaces

mlava:
fix for right sidebar pinned windows

@baibhavbista


mlava:
thanks for alerting me to the coming change and telling how to address it!

PR #893, opened 2024-06-12
Update Crosswords

mlava:
finally fixed the guesses save issue

### DIFF
--- a/extensions/mlava/crosswords.json
+++ b/extensions/mlava/crosswords.json
@@ -5,6 +5,6 @@
   "tags": ["crossword"],
   "source_url": "https://github.com/mlava/crosswords",
   "source_repo": "https://github.com/mlava/crosswords.git",
-  "source_commit": "91ac4f816c15f581c6d3e3c2bd0b2fc04a159407",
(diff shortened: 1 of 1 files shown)


baibhavbista (changes requested):
Loom showing the issue & reproduction: https://www.loom.com/share/15d699f88e4f4dbd9973e4265cb55aaf

not sure if this is just me or even if it is a new bug but I do not see the crossword in some cases
Have had it happen twice in quick succession
when this happens, a refresh or closing and opening parent again fixes it


image

mlava:
It looks like this issue is caused by a bug in [roamjs-components/components/ComponentContainer](https://github.com/RoamJS/roamjs-components/blob/main/src/components/ComponentContainer.tsx)

@mdroidian is trying to figure out whether it can be fixed. I'll let you know...

mlava:
@baibhavbista 
this was an issue in roamjs-components that has now been fixed

confirmed the render issue is now resolved

baibhavbista:
sorry for the late check @mlava 

lgtm will release

PR #894, opened 2024-06-14
Update Wordle

mlava:
Fix for broken answer source

### DIFF
--- a/extensions/mlava/wordle.json
+++ b/extensions/mlava/wordle.json
@@ -5,6 +5,6 @@
   "tags": ["wordle", "words", "games", "learning"],
   "source_url": "https://github.com/mlava/wordle",
   "source_repo": "https://github.com/mlava/wordle.git",
-  "source_commit": "f4fea907bb29ffef41f95dc8721c13fc54f250bb",
+  "source_commit": "92c1c92c48c6ed43064897a2938c24972ddf5ce5",
   "stripe_account": "acct_1MI2hiH4zkcxPFzj"
 }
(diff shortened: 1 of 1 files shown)


baibhavbista:
@mlava can you give more details on this new API you're hitting?
image

baibhavbista:
@mlava 
We talked about this in Slack, but adding here too for future reference:

googled a bit, and some people say that this website works: https://www.nytimes.com/svc/wordle/v2/{YYYY}-{MM}-{DD}.json

For today’s date, for example: https://www.nytimes.com/svc/wordle/v2/2024-06-25.json
`{"id":1791,"solution":"savor","print_date":"2024-06-25","days_since_launch":1102,"editor":"Tracy Bennett"}`
which seems to correspond to the answer when using your extension
image

If possible, we should try to use this API because:
1. is a NYTimes API so will probably be safe
2. will not require you to keep something running in heroku (which you were doing for the herokuapp.com server call above)

baibhavbista:
@mlava 
different behavior now using wordle extension
I think this could be fixed by using `response_data.days_since_launch` instead of `response_data.id`. I’m unsure why there are two different integer counts though
See sample json below:
{"id":1791,"solution":"savor","print_date":"2024-06-25","days_since_launch":1102,"editor":"Tracy Bennett"}
image

mlava:
> @mlava different behavior now using wordle extension I think this could be fixed by using `response_data.days_since_launch` instead of `response_data.id`. I’m unsure why there are two different integer counts though See sample json below: {"id":1791,"solution":"savor","print_date":"2024-06-25","days_since_launch":1102,"editor":"Tracy Bennett"} image

I'm really sorry. Fixed in latest merge

PR #897, opened 2024-06-14
Update: Full calendar

fbgallet:
@panterarocks49 in the mean time, i've added the possibility to add tags directly from the multiselect input !


fbgallet:
@panterarocks49 oh, just tested the published version, and click on checkbox to switch between TODO/DONE is broken, i've to fix it for v.2 !

fbgallet:
It's fixed

PR #904, opened 2024-06-25
Update Crosswords

mlava:
fixed guess storage issue (?finally)

sorry @baibhavbista but I think this one finally gets it done


baibhavbista:
@mlava 🤞 that this gets it done 😅

PR #910, opened 2024-07-17
Critical update: Full calendar

fbgallet:


baibhavbista:
@fbgallet could you mention what things this PR is meant to solve? (asking because the PR is named critical update so I assume there are critical bug fixes rather than just new features)

also I found a bug when I drag to a page that does not exist
image
I think the issue is the following: fn is supposed to return page title but at the end you call a fn which returns page uid
image

baibhavbista:
Another thing, I do not think you should pass uid here
In Roam, in some edge cases or for old DNPs, they might not have uids in the date format. But what you can be sure of is that the title should be in the proper format. In other words, we might have a page called "July 19th, 2024" in the daily log with a uid "dsadYdDf" (i.e. not "07-19-2024")

image

fbgallet:
@baibhavbista sorry for the inconvenience, there was a lot of changes in this PR, I should have totally indicated where the critical bug was fixed!

The main fix, as you yourself pointed out in your last comment, was to not provide uid when creating a new DNP, so I have added this function in roamAPI.js, to differentiate between creation of DNP and other pages:

image

I've also fixed the same bug when an event is dragged, thanks a lot for reporting it ! I have checked all possible actions (creating an event, internal and external drag and drop, resizing an event...) and there are no more issues now.

PR #921, opened 2024-08-16
Live AI Assistant: Important fixes + Groq support + direct requests to Anthropic API

fbgallet:


fbgallet:
I've fixed a weird bug that was only occurring on Safari,
and made a way better parsing of hierarchy in AI response.

fbgallet:
Fixed an issue on Post-processing
Added support to Groq (text generation models and audio transcription)

fbgallet:
Now support client-side Anthropic requests (my own backend on a distant server is no more needed)

PR #936, opened 2024-10-22
Full calendar: small update

fbgallet:


baibhavbista:
Correct me if I'm wrong @fbgallet , but it looks like you have some WIP code for google calendar but you commented them out while making this PR because you want users to be able to use the other stuff/fixes in the meantime?
(pretty sure this is the case but just checking)

Another thing, I saw that you have drag and drop new. I tried doing alt-drag of a block from Roam to the calendar and it created a new copy of the block instead of a block ref (which was my assumption, and is the default behavior when alt-dragging in Roam). Did you have any specific reason for doing so? 
(Another confusion is why dragging a block does not move the block but rather creates a copy. Is creating copies a common occurrence in some calendar workflows?)

fbgallet:
You're right, the code for Google calendar is only WIP, I should have created a branch for that !

The new drag&drop feature is about tags in the filter bar (you can now drag one of them in the calendar to create a new event with the corresponding tag). Dragging blocks into the calendar is an older feature. But you're totally right, it should be consistent with default Roam behavior, i will update it soon, I'll keep you posted.

fbgallet:
@baibhavbista i've made the small update, now Alt key create a block reference and simple dragging move the block !

PR #944, opened 2024-11-10
Live AI Assistant: SmartBlocks

fbgallet:
@baibhavbista, this is a test version, it's not to be merged for now

### DIFF
--- a/extensions/fbgallet/roam-extension-speech-to-roam.json
+++ b/extensions/fbgallet/roam-extension-speech-to-roam.json
@@ -5,6 +5,6 @@
   "tags": ["speech-to-text", "voice", "transcription", "recognition", "translation", "AI", "GPT", "Whisper", "Claude", "chat", "post-processing", "template", "context"],
   "source_url": "https://github.com/fbgallet/roam-extension-speech-to-roam",
   "source_repo": "https://github.com/fbgallet/roam-extension-speech-to-roam.git",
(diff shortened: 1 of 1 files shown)


fbgallet:
@abhayprasanna there is not much changes in the code, the SmartBlocks are mostly relying on existing code with a few tweaks to support option to limit captured levels in context.

The main additions are the SmartBlocks definition at the end of `./src/roamExtensionCommands.js` and the two functions at the end of `./src/utils.js`:
- `getContextFromSbCommand`, only relying on existing functions to convert pages or blocks content in string. The main updates are in the way parameters are parsed to define a very specific context at atomic level (with lists of block UIDs or page titles), which required enrich `getRoamContextFromPrompt` function with `getArgumentFromOption` function, also in `./src/utils.js`.
- `cleanFlagFromBlocks`, applied on children blocks when using <%LIVEAITEMPLATE%> Sb command or templated post-processing to remove `{text}` flag (which shows that the block containing this flag should not be altered or completed by the AI model but kept unchanged). This function rely on `roamAlphaAPI.updateBlock` API command, but it's use is very limited.

fbgallet:
@baibhavbista this PR is ready for review !

baibhavbista:
lgtm @fbgallet, thanks, will merge!
(I have a small question which I've messaged to you on Slack)

PR #947, opened 2024-11-18
feat: optimize URL parsing and prevent duplicate processing

dragonforce2010:
Add markdown link detection to avoid re-processing Improve URL replacement logic with negative lookbehind/lookahead Add content validation before processing URLs
Enhance error handling and task tracking
Fix issue with duplicate URL processing in same block

### DIFF
--- a/extensions/dragonforce2010/roam-website-title-parser.json
+++ b/extensions/dragonforce2010/roam-website-title-parser.json
@@ -5,6 +5,6 @@
   "tags": ["markdown"],
   "source_url": "https://github.com/dragonforce2010/roam-website-title-parser",
   "source_repo": "https://github.com/dragonforce2010/roam-website-title-parser.git",
-  "source_commit": "f36d584ab198c45a29af446a03c79a14489e8840",
+  "source_commit": "c22bc2e3f5ce59d10293de23b0a18502ac22765b",
   "stripe_account": "acct_1MI7MVQVLQeWv9Xo"
 }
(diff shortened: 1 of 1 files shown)


baibhavbista:
Hey @dragonforce2010 

(have also DMd you in Roam Slack about this)

We decided to host our own deploy of [CORS-anywhere](https://github.com/Rob--W/cors-anywhere) to make the experience better for extension developers. This way you wouldn’t have to have your own server for this extension.

Can you change your extension so that it calls this proxy url instead?

Some code generated by ChatGPT which I was able to use to get it work in your extension
```
async function getWebsiteTitle(url: string) {
  try {
      // Fetch the HTML content of the URL
      const response = await fetch(`[https://us-central1-firescript-577a2.cloudfunctions.net/proxy-corsAnywhere/${url}](https://us-central1-firescript-577a2.cloudfunctions.net/proxy-corsAnywhere/$%7Burl%7D)`);
      if (!response.ok) {
          throw new Error(`Failed to fetch: ${response.statusText}`);
      }
      
      // Get the text content of the response
      const html = await response.text();
      
      // Parse the HTML using DOMParser
      const parser = new DOMParser();
      const doc = parser.parseFromString(html, "text/html");
      
      // Get the <title> tag content
      const title = doc.querySelector('title').innerText;
      return title;
  } catch (error) {
      console.error("Error fetching the website title:", error);
      return null;
  }
}
```

Should be a pretty small change
Will finish up the review after this change

baibhavbista:
lgtm, will merge, thanks @dragonforce2010 !

PR #956, opened 2024-12-04
Update code-block-copy.json

8bitgentleman:
A regex fix and some visual feedback for copying


baibhavbista:
Hey @8bitgentleman I will merge this, but please do take a look at the slight bug I encountered below:

when you open a search window in the right sidebar for something like "```", then after a few results, it stops showing the copy button. See the one to the right in the screenshot below:
image
TBH I have no idea why this could be happening, if you open the same block in the main window it shows the button properly

PR #959, opened 2024-12-06
New Extension: Bluesky Extractor

8bitgentleman:
Extract Bluesky posts into your Roam Research graph with formatted text and images.


Stvad:
Ohh, nice, would love to see this in Depot!

PR #961, opened 2024-12-12
Update Wordle

mlava:
Uses the RR corsAnywhere instance rather than a heroku server to get the answer word. Faster


baibhavbista:
lgtm also worked when testing

PR #962, opened 2024-12-18
Roam-Reference-Expand: perf up

dive2Pro:
before: 
image
after:
image


baibhavbista:
Hey @dive2Pro , 

some comments:

**1. README images not working**

you can check via `(?)`
image

changing the readme like this would make it work I think
image


2. I've also DM'd you in Slack a loom showing slowdowns when using the extension when doing things like opening the sidebar. Can you check in a graph which has a lot of data in [[roam/reference extends/config]] and see if you can reproduce this on the old version and if it's better on the new version?

baibhavbista (changes requested):
https://github.com/Roam-Research/roam-depot/pull/962#issuecomment-2567348517

dive2Pro:
@baibhavbista 
1. Fixed.
2. Yes, I can reproduce this issue in the old version. I confirm that this problem should not occur in the newer version.

baibhavbista:
awesome @dive2Pro 
Thanks, I will release
btw I encountered a bug(?) where this extension does not seem to work with refs of blocks containing `>` I think? Can you check? https://www.loom.com/share/450ae7ce425949c7a943ab7c909b4277

PR #963, opened 2024-12-20
New Extension - API.Bible

mlava:


baibhavbista (changes requested):
Hey @mlava your extension doesn't work 😅. I think it might be because of a missing `build.sh`?

this is also why it's a good idea to test the PR-shorthand dev link 🤣

image

mlava:
@baibhavbista sorry, that's a noob mistake
Updated.

mlava:
Ok, @baibhavbista I think this is now ready for release. Thanks for the feedback on the earlier version :)
I decided not to use defaults but to prompt the user when they haven't completed the configuration.
I also found a bunch of issues with the external API data I was getting not being consistently formatted, and I hope I've handled them all now.
Let me know how you go!

PR #964, opened 2024-12-23
Add Create a new Block on a Daily page Roam Shortcut extension

Stvad:
This PR adds a new extension [Create a new Block on a Daily page Roam Shortcut](https://github.com/Stvad/roam-new-daypage-block).

### DIFF
--- a/extensions/Stvad/roam-date.json
+++ b/extensions/Stvad/roam-date.json
@@ -12,5 +12,5 @@
     "source_url": "https://github.com/Stvad/roam-date",
     "source_repo": "https://github.com/Stvad/roam-date.git",
     "source_commit": "e6fe27525aa5d3dad7398464f62f5f9c1c98efc4",
-    "stripe_account": "acct_1LcFfTHIh2HKWpPF"
+    "stripe_account": "acct_1LcbYoQmU9n2q3gV"
 }
--- a/extensions/Stvad/roam-date.json
+++ b/extensions/Stvad/roam-date.json
@@ -1,6 +1,6 @@
 {
     "name": "Date UX improvements",
-    "short_description": "Introduces UX improvements for interacting with Dates. Including a widget to easily change dates on mobile, natural language date creation and easier daily notes navigation",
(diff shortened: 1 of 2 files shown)


baibhavbista:
Hey @Stvad some comments:

1. the README.md for the extension is visible to users (in the help menu or when installing the extension in Roam Depot Marketplace), could you make the readme clearer considering that? For example, starting with "Roam Research Extension:" is not required)
image

2. I really like this extension, feels like such a simple solution to a common problem! But named as it is right now (i.e. "Create a new Block on a Daily page Roam Shortcut"), I would not be able to find it. Have you considered other names? Some names ChatGPT suggested "Capture on the Fly", "Jot in the Sidebar", "Quick thoughts", "Quick random thoughts", "Capture random thought"

3. More on 2, I would add some tags like "quick capture", "new note" in the extension json file. I would also change the "short_description" to be something like "Allows a quick-capture experience without having to change context of whatever you're working on. Just press a hotkey and start typing!"

4. Instead of your bespoke solution for hotkeys, can you use [extensionAPI.ui.commandPalette.addCommand](https://roamresearch.com/#/app/developer-documentation/page/XThfc9LJ6)? That way, users can also change the hotkey if they want a different one. (you want to be sure to pass the `default-hotkey` prop)

5. More re: hotkeys, I see that the hotkey is `ctrl-shift-n`. This is fine for mac, but pretty sure it tries to open new private window on Windows/Linux. Can you have a different one for windows/linux?
  i. I think `alt-n` might be a good one (corresponds to "Open daily notes" which is `ctrl-shift-d` in Mac and `alt-d` in windows/linux)
  ii. you probably want to do `window.roamAlphaAPI.platform.isPC ? "alt-n" : "ctrl-shift-n"` as the value to `default-hotkey`

baibhavbista (changes requested):
hey @Stvad I have a bunch of suggestions:
https://github.com/Roam-Research/roam-depot/pull/964#issuecomment-2567524637

panterarocks49:
Closing since stale, reopen if you wish to resubmit

PR #954, opened 2024-11-26
New extension: Last Year Today

qcrao:

PR #969, opened 2025-01-02
New extension: Last Year Today

qcrao:


qcrao:
@baibhavbista 

Thank you for the review! I’ve closed the original [MR](https://github.com/Roam-Research/roam-depot/pull/954) and opened a new one. In the updated MR, I’ve removed the logic that automatically opens the history page on load. Now, the history page will only open at the specified time. Let me know if you have any further feedback!

baibhavbista (changes requested):
Hey @qcrao sorry, it looks like I forgot to do a detailed review the last time around. 

I found some issues: 

**1. the setting for time does not seem to work** (probably because you're using diff names for reading and writing, maybe a refactor gone wrong? - "daily-update-hour" vs "hour-to-open-last-year-today-page")
image

**2. there are a lot of unused dependencies, could you clean those up?** (probably from init code for the extension itself)
image

3. your solution to the problem of historical pages seems to be to only show that at the exact time. As you probably know yourself, issue with this is that if the user opens the graph at 9:01 am, they won't see the historical pages. 
Might I suggest a different solution? When you open the historical pages, you do something like `extensionAPI.settings.set("last-opened-day", "01-07-2025");`. And when you load the extension, you first check this setting. If value is same as today, do nothing, otherwise open the historical pages

qcrao:
> Hey @qcrao sorry, it looks like I forgot to do a detailed review the last time around.
> 
> I found some issues:
> 
> **1. the setting for time does not seem to work** (probably because you're using diff names for reading and writing, maybe a refactor gone wrong? - "daily-update-hour" vs "hour-to-open-last-year-today-page") image
> 
> **2. there are a lot of unused dependencies, could you clean those up?** (probably from init code for the extension itself) image
> 
> 3. your solution to the problem of historical pages seems to be to only show that at the exact time. As you probably know yourself, issue with this is that if the user opens the graph at 9:01 am, they won't see the historical pages.
>    Might I suggest a different solution? When you open the historical pages, you do something like `extensionAPI.settings.set("last-opened-day", "01-07-2025");`. And when you load the extension, you first check this setting. If value is same as today, do nothing, otherwise open the historical pages

@baibhavbista 

Thank you for the detailed review. Here's my response:

1. The hours setting issue has been fixed. The configuration key names have been standardized for consistency.

2. I've cleaned up the unused dependencies.

3. Thank you for the suggestion regarding the history pages implementation. I've modified the logic:
   - Users will see 3 history pages (configurable) immediately upon loading the extension
   - After initial load, pages will open based on the configured hour (defaults to 9 AM)
   - If the hour setting isn't configured, it uses the default value
   - This works well for both usage patterns:
     - Users who close Roam daily will see the history pages upon reopening
     - Users who keep Roam running will see pages at the configured time
   - Additionally, I've added keyboard shortcuts and commands for quickly opening/closing history pages

Regarding the testing issue with qcrao+last-year-today+969: I noticed that extension.js might not be updating correctly during testing. The local file works fine, but there might be caching issues in the test environment. One way to verify this is to check the default history page count in settings - it should now show 3 instead of the previous default of 1.

Could you please verify if you're experiencing similar testing issues on your end?

Please let me know if there are any other issues you'd like me to address.

image

baibhavbista:
@qcrao 
Thanks for letting us know about the extension.js not getting built correctly. Our best guess right now is that your update/builds happened right as we were making some changes to the build scripts, and that is why the build silently failed

I just reran the build now and it seems to be working with the latest version
image


btw I have a comment regarding 3
image
The usecase I am unsure of if you're considering here is "user who opens and closes their graph multiple times a day". Currently the history pages would open every time, is this intentional? My earlier suggestion of `extensionAPI.settings.set("last-opened-day", "01-07-2025");` was partly to handle this use case btw

However, this PR has been open for quite a long time without release, so I will release it. If you want to address the above comment, please do so and open up a new PR

qcrao:
> @qcrao Thanks for letting us know about the extension.js not getting built correctly. Our best guess right now is that your update/builds happened right as we were making some changes to the build scripts, and that is why the build silently failed
> 
> I just reran the build now and it seems to be working with the latest version image
> 
> btw I have a comment regarding 3 image The usecase I am unsure of if you're considering here is "user who opens and closes their graph multiple times a day". Currently the history pages would open every time, is this intentional? My earlier suggestion of `extensionAPI.settings.set("last-opened-day", "01-07-2025");` was partly to handle this use case btw
> 
> However, this PR has been open for quite a long time without release, so I will release it. If you want to address the above comment, please do so and open up a new PR

Thank you for merging this MR!

Regarding comment 3, I believe most users don’t frequently open and close Roam multiple times a day, so this scenario might not be a common problem. For now, I’ll keep the current behavior as is. If I receive user feedback about this in the future, I’ll consider adding a toggle to limit opening the history pages to only once per day.

PR #970, opened 2025-01-02
Update export-table to support :q tables

8bitgentleman:


baibhavbista:
looks good @8bitgentleman , thanks!
Will release!

PR #987, opened 2025-02-08
New Plugin: Roam Recents

dive2Pro:
image

this plugin will provide users with a view to review the pages they have recently visited.


dive2Pro:
@panterarocks49 thanks, all set with the fields

panterarocks49 (approved):

PR #988, opened 2025-02-14
Update todo-progress-bar.json

8bitgentleman:
Adds the ability to pass custom text into the component, instead of just the standard "Done" text. Also refactored to add a settings panel instead of relying on editing the component text parameters directly

### DIFF
--- a/extensions/8bitgentleman/todo-progress-bar.json
+++ b/extensions/8bitgentleman/todo-progress-bar.json
@@ -5,6 +5,6 @@
     "tags": ["roam/render", "TODO", "CSS"],
     "source_url": "https://github.com/8bitgentleman/roam-depot-todo-progress-bar",
     "source_repo": "https://github.com/8bitgentleman/roam-depot-todo-progress-bar.git",
-    "source_commit": "13755f6d079895c42ade68957e16e70889278e34",
+    "source_commit": "470fe2cc6dfba309183d7bb7fe3ea28269011fcd",
     "stripe_account": "acct_1LJFEYQmxalymEZL"
   }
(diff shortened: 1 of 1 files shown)


panterarocks49:
Hey I installed this on a graph which has v11 installed but it doesn't seem to be updating the last version, is the update mechanism working for you?

I also tried it on a graph which didn't have any todo progress bar installed before, I installed the version on the store, then uninstalled it, then I installed this PR and it didn't update

panterarocks49:
@8bitgentleman just checking in, did you see this? (saw you posted some other updates since)

8bitgentleman:
I totally missed this, thanks for the ping. I'll take a look today

8bitgentleman:
@panterarocks49  Looks like with the move to using a global window object the actual code block updating was lost. I've added functions in for this so the code block should update now when there's a new version

panterarocks49:
I think there might be a bug with embed counting
This is with embeds on, it double counts the embed
image


with embed counting off, it is still counting the embed, but now it counts it correctly
image

8bitgentleman:
@panterarocks49 what's the string in that block? Does the embed string have a weird format ? 
I've generalized the regex but even before I wasn't able to reproduce what you're showing. Can you share your block structure?

panterarocks49:
I deleted the original block but this is similar 
image

also noticed that the version doesn't update here
image

panterarocks49:
some of the logging
image

8bitgentleman:
@panterarocks49  I think counting TODOs in embed children is going to take some more time to cook. For now I've removed that bit so we can merge the other improvements. As for the template title I kind of regret putting that in. I've now got no way to actually access and update it..

PR #993, opened 2025-02-21
Live ai assistant - small but critical fix (currently the extension isn't loaded)

fbgallet:
Strangely, Toaster module was not imported in another module and it was not a pb in dev, but it prevents the extension to load in production!
It was just a line of code, it's fixed


fbgallet:
@panterarocks49 sorry, the extension is not loading after update, there was a small hidden bug, it's fixed here !

PR #1014, opened 2025-03-24
New extension: URL to Selection

qcrao:


panterarocks49:
Hi, I can't accept this extension because this is something which Roam already does. We already detect if you are pasting a url and we convert the selected text into a markdown link. Additionally Roam detects if you are pasting a block ref or page ref and converts those to markdown links if there is selected text

Is there a bug with our implementation? why did you make this?

qcrao:
Oops, I didn’t notice that before — you’re right, it already works great. Thanks!

PR #1019, opened 2025-04-01
Update Page History

dive2Pro:
- Rename to Roam History
- Remove sync data feature. All data is now saved locally
- Add block timeline feature


panterarocks49:
Why remove the sync feature? If I use Roam across devices then the history won't be available on both

Your settings tab title is still "Page History" https://github.com/dive2Pro/roam-snapshot/blob/main/src/config.ts#L20

dive2Pro:
@panterarocks49 yeah, your concern is totally normal, which prompted me to bring this feature back. You can check out the details in the ReadMe under the `AutoSync` section. Also, the text in settings has been updated.

panterarocks49 (approved):
Thanks!

PR #1029, opened 2025-05-07
New Extension: Image Tools

qcrao:
Mobile-Friendly Image Enhancements for Roam

This PR introduces a new extension: Roam Image Tools, focused on improving image interaction in Roam, especially on mobile devices.

Key Features:
- 👁️ Image Zoom Modal:
Tap the eye icon to view images in fullscreen — ideal for mobile users who can’t click to enlarge images like on desktop.
- 📋 Copy to Clipboard:
One-tap copy of images to the clipboard, no right-click needed.
- 💾 Quick Save:
Easily download images with a single tap.

These tools appear on all images and are designed to enhance usability across devices, particularly where native image interaction is limited (e.g., touch devices).

### DIFF
--- a/extensions/qcrao/image-tools.json
+++ b/extensions/qcrao/image-tools.json
@@ -1,0 +1,10 @@
+{
+  "name": "Image Tools",
+  "short_description": "Enhance your image experience in Roam Research",
+  "author": "qcrao",
+  "tags": ["productivity"],
+  "source_url": "https://github.com/qcrao/image-tools",
+  "source_repo": "https://github.com/qcrao/image-tools.git",
+  "source_commit": "e0783651fe9e8f368b275428e6df865060fd35d3",
+  "stripe_account": "acct_1PJTCpQd6q6JNEre"
+}


panterarocks49:
Instead of fetching the image src, you should use our file API to fetch the file object https://roamresearch.com/#/app/developer-documentation/page/Rf-zubF-I. This way encrypted images will work.

I am also confused why you need to convert the img element to a canvas to write to the clipboard? Is that an optimization to not refetch the image?

Every time I click the copy button I get the error "ClipboardItem failed, trying fallback method NotAllowedError: Failed to execute 'write' on 'Clipboard': Type image/jpeg not supported on write." and then it copies the image url. When I click the copy button on mobile, it opens a popup in a new window (first chrome actually blocks the popup, I had to allow it). I'm not sure it copies anything on mobile. 

I also get some buggy behavior with the download button on mobile, it opens the share menu, but if I click out of the share menu it opens a popup. 

https://github.com/qcrao/image-tools/blob/e0783651fe9e8f368b275428e6df865060fd35d3/src/index.tsx#L48
Is this command necessary? It won't be clear to the users what this does

PR #1056, opened 2025-08-05
Add TODO Analysis extension

forrestchang:
This PR adds the TODO Analysis extension to Roam Depot.

## Extension Details
- **Name**: TODO Analysis
- **Author**: Forrest Chang
- **Description**: Comprehensive analytics for TODO items with charts, achievements, and productivity insights
- **Repository**: https://github.com/forrestchang/roam-todo-analysis

## Features
- Task completion analytics and trends
- Productivity score calculation
- Achievement system with 70+ unlockable achievements
- Interactive charts and heatmap calendar
- Daily logbook view
- Gamification with levels and XP

The extension helps users track their task completion patterns and improve productivity through data-driven insights.

### DIFF
--- a/extensions/forrestchang/roam-todo-analysis.json
+++ b/extensions/forrestchang/roam-todo-analysis.json
@@ -1,0 +1,9 @@
+{
+  "name": "TODO Analysis", 
+  "short_description": "Analyze TODO blocks in your Roam graph, providing insights and statistics on task patterns and completion rates.",
+  "author": "Forrest Chang",
+  "tags": ["todo", "statistics", "analysis", "productivity"],
+  "source_url": "https://github.com/forrestchang/roam-todo-analysis",
+  "source_repo": "https://github.com/forrestchang/roam-todo-analysis.git",
+  "source_commit": "c317f3cd57fc90afc681d78c6951d786a7f31b90"
+}


panterarocks49:
Hi thanks for submitting your first extension!

While not an issue directly, why did you not use react? I think this would have been a lot simpler if you had used react, but at this point I wouldn't change it. 

The issue that stems from that is your use of innerHtml can be subject to XSS attacks since you are displaying blocks / pages in it. For example, if you create a block like this

```
{{[[TODO]]}} <img src=x onerror=alert(1)>
```

When you open the todo manager you will get an alert. This can be used to run arbitrary javascript, and since extensions can be enabled by admins for other users, this is an attack vector. 

Here is the list of problem areas I found

- **Logbook → task cards**
- 
  Inside `createLogbookView` → `updateLogbook()` → `tasks.forEach(...)`:
- ```javascript
  taskEl.innerHTML = `
    ...
    <div style="color: #182026; margin-bottom: 6px; line-height: 1.4;">${displayContent}</div>
    ...
    <span ...>📄</span> ${task.pageTitle}
    ...
  `;
  ```
    - `displayContent` comes from `task.content` → **block text** (with highlight applied).
    - `${task.pageTitle}` is the **page title** from Roam.
- **Search → page headers**
- 
  Inside `createSearchView` → `displaySearchResults()`:
- ```javascript
  pageHeader.innerHTML = `
    <span>📄 ${pageTitle}</span>
    <span ...>${pageTasks.length} task${...}</span>
  `;
  ```
    - `${pageTitle}` is the **page title** (user-controlled).
- **Search → task rows**
- 
  Inside `createSearchView` → `displaySearchResults()` → `pageTasks.forEach(...)`:
- ```javascript
  taskEl.innerHTML = `
    ...
    <div style="color: #182026; line-height: 1.5;">${displayContent}</div>
    ...
  `;
  ```
    - `displayContent` is derived from `task.content` → **block text**.

The safest thing to do here is to convert this to code creating elements directly with `.textContent` (claude can probably do this easily for you)

---

I also need you to create a file called `build.sh` in the base directory, inside it will run the `node build.js` to build the extension, and you can delete the `extension.js`. We build it in github actions to make sure it hasn't been tampered with.

---
This you don't need to change but you can speed this [todo search query](https://github.com/forrestchang/roam-todo-analysis/blob/1edceb308033de069ee5f3da332697b6be47bc18/src/queries.js#L17) up by first looking for the reverse :block/refs of the status page. I didn't test it but something like this should work to cut the query time down
```
[:find ?uid ?string ?create-time ?edit-time ?page-title
 :where
 [?status-id :node/title "${status}"]
 [?b :block/refs ?status-id]
 [?b :block/uid ?uid]
 [?b :block/string ?string]
 [(clojure.string/includes? ?string "{{[[${status}]]}}")]
 [?b :block/page ?page]
 [?page :node/title ?page-title]
 (or-join [?b ?create-time]
   (and [?b :create/time ?create-time])
   (and [(missing? $ ?b :create/time)]
    [(ground 0) ?create-time]))
 (or-join [?b ?edit-time]
   (and [?b :edit/time ?edit-time])
   (and [(missing? $ ?b :edit/time)]
    [(ground 0) ?edit-time]))]
```

The same goes for the other queries you have in that file, searching the backrefs should be much faster

I think you could even cut out the ` [(clojure.string/includes? ?string "{{[[${status}]]}}")]` too at that point since if something is marked as [[TODO]] or [[DONE]] but doesn't have the {{ it probably means the same thing.

PR #1064, opened 2025-09-03
New extension: Roam Copilot

qcrao:
🚀 **Transform your Roam Research experience with AI-powered insights**

Roam Copilot brings the power of advanced AI directly into your Roam workspace. Get instant, context-aware assistance for brainstorming, analysis, and idea connections without ever leaving your graph.

[Watch the video](https://youtu.be/SaAvypfKgUo)
👆 [Click to watch the demo video](https://youtu.be/SaAvypfKgUo)

### DIFF
--- a/extensions/qcrao/roam-copilot.json
+++ b/extensions/qcrao/roam-copilot.json
@@ -1,0 +1,10 @@
+{
+  "name": "Roam Copilot",
+  "short_description": "Your AI co-pilot in Roam.",
+  "author": "qcrao",
+  "tags": ["productivity", "AI", "copilot"],
+  "source_url": "https://github.com/qcrao/copilot",
+  "source_repo": "https://github.com/qcrao/copilot.git",
+  "source_commit": "0ed18b2a7e8cba7cf71a4a40a4ff87957a901755",
+  "stripe_account": "acct_1PJTCpQd6q6JNEre"
+}


panterarocks49:
This is very cool! I was thinking of building a similar integration with chat

How do you determine what context to send to the LLM? I noticed that I can reference pages and drag blocks in, but what confuses me is that sometimes the current view I'm looking at shows up like this

image

But then when I ask a question that current open page isn't included in the context. 

I think I also ran into a bug where a past conversation was continued into a new one, the convo on the right is scrolled to the top, but see the network request contains messages from the last conversation I had

image


---

What is going on with the build.sh script? it looks like it builds the extension twice and once in dev mode? is that intentional? Can you explain what this is doing?

image

panterarocks49:
Now I'm getting huge spikes in CPU usage when I first load the (large) graph. I think it gets stuck in a loop for a bit, but in this case it did get out of the loop. However when I closed the sidebar and then reopened it it froze the page entirely

image

I also see this when typing in the chat window, I think it's trying to refetch the context too often and on this large graph it freezes the UI. 
Other than reducing the frequency of refetching the context, which I also think you should do, I think you should switch from using queries to the pull API. The pull API is much much faster

An example would be here, this code must run very slow
https://github.com/qcrao/copilot/blob/main/src/utils/roamQuery.ts#L166
```
[:find ?uid ?string ?order
         :where
         [?parent :block/uid "${blockUid}"]
         [?parent :block/children ?child]
         [?child :block/uid ?uid]
         [?child :block/string ?string]
         [?child :block/order ?order]]
```

Instead you can do a pull, with a depth limit, and then traverse the output

```
[:block/uid
:block/string
:block/order
{:block/children <insert-depth-here>}]
```

qcrao:
@panterarocks49  Thanks for calling all of this out. 

I’ve switched our block lookups over to a single pull call with a bounded depth instead of the repeated q queries, which gets rid of the slow loop you pointed out in roamQuery.ts. 

On top of that I tightened the context-refresh logic: every trigger shares the same throttle guard, the daily-note view no longer spins, and navigation still refreshes quickly, so the sidebar reopen/typing scenarios no longer slam the CPU. 

Let me know if anything still feels off on your big graph.

qcrao:
@panterarocks49  Thank you for the detailed feedback! I really appreciate you taking the time to test and report these issues. Here are my updates:

1. Performance Improvements

I’ve implemented several optimizations to address the typing lag:
- Optimized chat input state management: the input now keeps its own serialized state with a minimal ref API, avoiding parent re-renders on every keystroke.
- Streamlined widget state: removed redundant inputValue state from CopilotWidget to prevent unnecessary re-renders during clearing, template inserts, and date substitutions.
-Reduced console noise: moved verbose logging (including emoji-heavy debug output) behind VERBOSE_* switches.

Please let me know if you still notice any performance issues.

2. Context Window Leakage

I couldn’t reproduce this issue locally.

Please try restarting Roam Research. Some changes require a full restart to take effect. If the issue persists, I’ll dig deeper.

- At 5:27, the screenshot shows one conversation with its context.
- At 5:28, a new conversation starts, and it does not include the context from 5:27.

<p align="center">
  
  
</p>


3. Page Context Depth

Fixed ✅ The system now fetches complete nested block structures for both page content and backlinks, not just the first level. This should provide full context depth for better reasoning.

---

Thanks again for the helpful review! Please test these changes and let me know if you run into anything else.

panterarocks49:
I am still getting the context issues, but the other issues seem solved now

Here is a loom showing how to reproduce it
https://www.loom.com/share/b793c93dc1df40afbdf95cc3f9989724

qcrao:
请问这个PR预计什么时候可以合并？目前还有什么问题或需要改进的地方吗？感谢！

panterarocks49:
also maybe the default models should be gpt 5? I tried entering "gpt-5" into the custom models and I get this error
image

qcrao:
@panterarocks49 Thanks for the feedback! The issue has been fixed, and the GPT-5 model is now available.
Please note, however, that GPT-5 currently requires an organization to be recognized; otherwise, it may return an error.

PR #1068, opened 2025-09-18
Live AI assistant - update

fbgallet:
It's a big update with total revamp of Ask your graph agent (in /src/ai/agents/search-agent ) handling multiple tools to convert the natural language request into multiple (and eventually chained or combined) datomic queries and supporting multiple level of semantic variations.
It's also introducing a MCP server agent.

Ask your graph has a very big codebase, sorry for that! Most Roam API call are in multiple executors.ts files in agent tools. The big and not very optimized part is /findBlocksWithHierarchy/ tool, combining multiple datomic queries to test if blocks and multiple level of children matche the conditions (for ex. a simple A and B request will be converted into A <=> B request, testing if A and B are in the same block, or A in a block with B in a child, or B in a block and B in a child...)

It's not ready to be published, not yet fully tested and the documentation has not yet been updated.


fbgallet:
@panterarocks49 I've added DOMPurify for innerHTML injections and updated readme (the full doc for Ask Your Graph agent can come later since it will be in a separate .md in /docs/ ), so it's almost ready for merging

PR #1285, opened 2026-03-07
New Extension: Table-plus

8bitgentleman:
A new extension that enhances roam native tables with direct cell editing, sorting, filtering, and basic excel formulas.

### DIFF
--- a/extensions/8bitgentleman/table-plus.json
+++ b/extensions/8bitgentleman/table-plus.json
@@ -1,0 +1,14 @@
+{
+    "name": "Table Plus",
+    "short_description": "Enhances roam native tables with direct cell editing, sorting, filtering, and basic excel formulas.",
+    "author": "Matt Vogel",
+    "tags": [
+      "tables",
+      "filters",
+      "sorting"
+    ],
+    "source_url": "https://github.com/8bitgentleman/roam-depot-tables-interop",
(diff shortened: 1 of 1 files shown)


8bitgentleman:
Hi Roam team, this extension is working  well overall but I've hit a crash that I think is related to the BlockString API usage

 _**What I'm doing:**_ Use `window.roamAlphaAPI.ui.react.BlockString` to render rich cell content
   (TODO toggles, block embeds, page links, etc.) inside a custom ReactDOM.createRoot container
   that is mounted alongside the native .roam-table element.

  _**The crash:**_ When a user interacts with interactive elements rendered by BlockString (mostly the block embed expand ) the full Roam tab crashes (DevTools disconnects) on block embed interactions. On TODO clicks it's a recoverable JS error.

My best guess is that BlockString's interactive handlers are wired into Roam's own React event system. When triggered, Roam's global block-editor click handler fires and tries to find block editor context that doesn't exist in our foreign React root, resulting in a null block string → .substring() crash.


Am I overcomplicating. this? Is there a supported way to render BlockString content inside a custom ReactDOM.createRoot such that its interactive handlers don't trigger Roam's block-editor click handler?

panterarocks49:
closing this for now since we have interactive tables

PR #1330, opened 2026-04-15
Add Roam Toolkit Vim Mode extension

Stvad:
## Summary
- add the Roam Toolkit Vim Mode extension metadata
- point roam-depot at the published source repo and commit c05674379bd1dbe938577b6b01b620df8ffb5514
- reuse existing Stvad marketplace author and stripe account metadata

### DIFF
--- a/extensions/Stvad/roam-vim-navigation.json
+++ b/extensions/Stvad/roam-vim-navigation.json
@@ -10,6 +10,6 @@
   ],
   "source_url": "https://github.com/Stvad/roam-vim-navigation",
   "source_repo": "https://github.com/Stvad/roam-vim-navigation.git",
-  "source_commit": "7ffd05938301588f2bc9de67fe66ed8c2224bdd8",
+  "source_commit": "923a29b913c4d7eb49c320f68098e478b9e8372c",
   "stripe_account": "acct_1LcbYoQmU9n2q3gV"
 }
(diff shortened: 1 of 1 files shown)


panterarocks49:
Awesome! There is one small thing to fix, dependencies that Roam exports to the window should be reused

LIke this in the webpack config

```
await build({
      bundle: true,
      entryPoints: ['src/ts/roam-vim-plugin/index.tsx'],
      format: 'esm',
      external: ['react', 'react-dom'],
      // ...
  })
  ```
  
  here is the list of deps https://roamresearch.com/#/app/developer-documentation/page/QBK1-Gh3J
  
 At least react and react-dom should be loaded from here, and ideally cytoscape as well but cytoscape you might have to use directly from the window to wait on the promise since it's loaded async

Stvad:
This is a repackaging of a fairly old codebase (react 16.2 and related dependencies that rely on it). I'm concerned that upgrading everything to 18.2 would not be straightforward =\

Stvad:
I'll try to make a coding agent do an upgrade and also clean up unused code/deps though. As vim side of things doesn't really do any ui work (though it relies on react-hotkeys for shortcut management)

panterarocks49:
oh well that's ok if you can't get it to work. we run react 18 in compat mode so I would be surprised if there were any significant changes you had to make.

Stvad:
Seems to work, ok!

Stvad:
I've done a bunch of updates to this in my local version - moving things that can be moved from browser simulation to roamAlphaApi and moving away from react-hotkeys to remove react dependency alltogether. 
That's quite a large number of changes, so testing that locally for now

Stvad:
Seems to be working well, updated a hash to the latest!

panterarocks49:
perfect, thank you!

PR #1371, opened 2026-07-05
Add Related Notes extension

qcrao:
## Related Notes

Semantic search and automatic related-note discovery, powered by Roam's native semantic engine (`semanticSearch` API).

- **Related Notes panel**: as you browse, a draggable/resizable panel automatically surfaces semantically similar pages and blocks — grouped by chunk with per-page diversity, so one long note never floods the list
- **Semantic search** (`Ctrl+Shift+S`): search the graph by meaning, keyboard-first; works cross-lingual (e.g. a Chinese query finds English notes)
- **Graceful fallback**: if the graph's semantic embeddings are off, everything still works via literal matching, with a hint on how to enable the engine
- **Strictly read-only, zero external requests, no telemetry**: the extension only calls Roam's built-in APIs; embedding routing is entirely controlled by the user's own Roam settings

Repo: https://github.com/qcrao/roam-related-notes (README includes a detailed privacy section and a technical deep-dive doc)

Author: qcrao (existing Depot extensions: Roam Copilot, Last Year Today, Card Theme, etc.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01NZheN6BbWr4zmE8jLiSSpg

### DIFF
--- a/extensions/qcrao/roam-related-notes.json
+++ b/extensions/qcrao/roam-related-notes.json
@@ -1,9 +1,14 @@
 {
   "name": "Related Notes",
-  "short_description": "Find notes you forgot you wrote. Semantic search and automatic related-note discovery, powered by Roam's native semantic engine.",
+  "short_description": "Find notes you forgot you wrote. As you browse, a panel automatically surfaces semantically related pages and blocks, powered by Roam's native semantic engine.",
   "author": "qcrao",
-  "tags": ["search", "ai", "navigation", "productivity"],
+  "tags": [
+    "search",
+    "ai",
+    "navigation",
+    "productivity"
+  ],
   "source_url": "https://github.com/qcrao/roam-related-notes",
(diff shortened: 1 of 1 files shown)


panterarocks49:
Do you think it's necessary to add another semantic search window? The default search has a semantic option already. seems like it might be confusing and our default window also has more filter options

I think it would be good to say somewhere in the readme that it takes the title and the first 10 direct children to compare

qcrao:
Thanks for the review — both points addressed:

1. **Removed the semantic search modal entirely.** You're right: now that the default search has a semantic option (with more filters), a second search window is redundant and potentially confusing. The extension is now a single surface — the Related Notes panel. (qcrao/roam-related-notes@07bf177)
2. **Added a "How it works" section to the README** explaining that the panel builds its query from the page title + the first 10 direct children (truncated to 600 characters), and updated the privacy section to reference it.

`source_commit` in the metadata is updated to pin the new commit.

PR #1385, opened 2026-07-10
similar-pages modernization

8bitgentleman:
The original similar-pages extension leaned on an old embedding stack and a scoring formula that fought itself. On a real graph it surfaced weak, noisy matches. This modernizes the semantic core and fixes races that could leave the panel stuck or showing stale results.

The changes fall into two buckets: **(1) a modernized similarity engine** and **(2) correctness hardening** around the async embedding pipeline.

1. Modernized similarity engine (three changes, each isolated + measured on a ~2,300-page graph):
- Model: USE → BGE-small-en-v1.5 - top-10 avg cosine 0.38 → 0.71. Still fully local via transformers.js.
- Content window: 250 → 2000 chars - old cap embedded ~title + first block; new window embeds what the page is actually about.
- Scoring: distance × similarity → similarity + 0.1 × distance - old formula structurally hid your most-similar pages (multiplying by distance zeroed out close matches); new one ranks by similarity with a small bonus for far-but-similar "discoveries."

2. Correctness hardening - partial-embedding race (generation guard), duplicate IndexedDB connection (shared ref), silent worker failures (error card instead of infinite spinner), dtype: "q8" API update.

(Permission was given by the original author for this fork)

### DIFF
--- a/extensions/phonetonote/similar-pages.json
+++ b/extensions/phonetonote/similar-pages.json
@@ -5,6 +5,6 @@
   "tags": ["machine learning", "natural language processing", "nlp", "graph theory", "universal sentence encoder"],
   "source_url": "https://github.com/8bitgentleman/roam-depot-similar-pages",
   "source_repo": "https://github.com/8bitgentleman/roam-depot-similar-pages.git",
-  "source_commit": "67f3e6b608bceb9e758b4790d544ae2394edbdf9",
+  "source_commit": "98171a71db25aa27eb9c9f0a5ab6fbf0e768afbe",
   "stripe_account": "acct_1LJFEYQmxalymEZL"
 }
--- a/extensions/phonetonote/similar-pages.json
+++ b/extensions/phonetonote/similar-pages.json
@@ -1,10 +1,10 @@
 {
   "name": "Similar Pages",
   "short_description": "uses machine learning and graph theory to help you find connections between ideas",
-  "author": "scott block",
+  "author": "scott block & Matt Vogel",
   "tags": ["machine learning", "natural language processing", "nlp", "graph theory", "universal sentence encoder"],
-  "source_url": "https://github.com/phonetonote/similar-pages",
-  "source_repo": "https://github.com/phonetonote/similar-pages",
-  "source_commit": "3ff405740863f73621cc0c7bccaee5909d2e685e",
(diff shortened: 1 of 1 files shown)


panterarocks49:
Am I doing something wrong? I installed the extension, hit the scatterplot button and selected a page and it's blank
image

8bitgentleman:
Is this an isolated page with nothing linking to it?Is this a particularly empty graph by any chance (looks like it's named 'test103')

panterarocks49:
Oh you're right I was on a semi-empty test graph, thought I was on a different one. Now I get results

image

I wouldn't say these pages are very related though, are you getting better results? 
For example if I use our semantic search, it's embedding + keyword so I scrolled past the keyword results. I'd expect datahike and datascript to be in the results

image

Do you get better results than this?

panterarocks49:
It would be nice if the search split the query and searched on individual words. if I type "project search" it misses "Project: Better Search". And sorting by shortest / bringing exact matches to the top. I searched for "backend" but the exact match was way down the list

8bitgentleman:
This is helpful, thanks. I've been testing in a densely connected graph so some of the shortcuts I took were masked
**Page picker:** queries are tokenized now and ranked so that should make it a bit better
**Similarity:** Some trade offs here. Previously the extension scored pages against others that were reachable through it(ie linked) and if the page was unlinked it fell back to relying on the title and the recently edited pages which isn't as useful as it seemed in retrospect.
This should be fixed by embedding the whole graph once (cached by content hash) and ranking the apex against the entire graph so graph distance is now a secondary "surprising discovery" signal instead of what decides who's a candidate. This should specifically repair the orphan/new-page case from above. The cost of this is a a slow first run while embedding the whole graph. I reworked the worker so the model loads once rather than per chunk after that

Let me know if you're getting any better results now?

panterarocks49:
These results look a lot better! 

image



Claude found out a possible bug
image

Also might be worth aligning the idb versions between the worker and main thread? Usually I would say use the version we expose but it wouldn't be available in the worker. 
image

8bitgentleman:
Glad the good results are coming through now! New commit should cover the bug claude found. For the idb version comment I dug into it but bumping that to match the worker fails with ERESOLVE because of roamjs-components peer-pins. I decided to leave it alone for now

PR #1398, opened 2026-07-25
Live AI: update

fbgallet:
- **New models support**: Claude Opus 5, Gemini 3.6 Flash & 3.5 Flash Lite
- **Attached files as input**, beyond `.pdf`: `.md`, `.txt`, `.csv`, `.json`, code files… are read and inserted in your request, so **any model** can use them (even local ones). `.docx`/`.pptx`/`.xlsx` work with **OpenAI models**.
- Fixed **Gemini ignoring a .pdf from the context** at medium/high thinking level, and repaired `.pdf` support for OpenRouter, Groq, DeepSeek, Grok, custom endpoints, and Claude models with thinking enabled.


panterarocks49:
Hi Fabrice — one thing to fix, then good to merge.

attachedFileRegex matches ordinary URLs as attached files (src/utils/regex.js:53-58)

"org" is in textFileExtensions (for org-mode files), and the bare-URL alternative lets the extension sit right after the domain:

https://en.wikipedia.org/wiki/Roam   →  matches "https://en.wikipedia.org"

The extension tries to download the domain root as a text file, CORS blocks it, and the user gets ⚠️ Attached file(s) that could not be read and will be ignored: document. — on every request with a .org link in the prompt or context. If CORS does allow it, the match is only the bare domain, so replacing it corrupts the URL: https://en.wikipedia.org/wiki/Roam → <page HTML>/wiki/Roam.

Same class: .php URLs, .sh/.rs/.md domains (all real TLDs), .js/.css asset URLs. And the first alternative keys off the markdown label, not the URL, so [CHANGELOG.md](https://github.com/foo/bar) matches and fetches the GitHub page.

Suggest requiring a path segment rather than allowing the extension immediately after the domain, and having the label agree with the URL instead of deciding on its own.

Minor

- An Office link in the context replaces the entire answer with the "can't read Office documents" warning on non-OpenAI models (multimodal-commands.ts:~917), so a stray .docx on the page blocks the chat. A toast plus continuing would be friendlier.
- inlineAttachedFiles has no dedup or cache and MAX_INLINED_FILE_CHARS is per-file with no overall cap, so the same URL in prompt and context is fetched twice and N links means N × 100k chars.
- The new model-updates.json entries omit the -latest aliases their modelRegistry.js counterparts have (opus-latest, gemini-flash-latest, gemini-lite-latest) — intentional?

Otherwise nice work — the pdfLinkRegex.lastIndex resets, the usesResponsesApi split, and the replacer function in inlineAttachedFiles are all good catches. No new deps, no dynamic code execution, no new endpoints, nothing touching block or page writes.

fbgallet:
Hi,
the regex is fixed now, the two minor issues too (the last one is intentional, model-updates.json is only temporary)

fbgallet:
Added support for `gpt-transcribe` and `gpt-live-transcribe` => new Live voice transcription feature

